PR #5, opened 2017-11-25
[DOXIA-534] Migrate logging to Sl4j

slachiewicz:
Please check. 

Current build fails with cirr error - due to interfaces change.


doxia-modules/doxia-module-fo/pom.xml
```diff
@@ -78,7 +72,7 @@ under the License.
     <dependency>
       <groupId>commons-collections</groupId>
       <artifactId>commons-collections</artifactId>
-      <version>3.2.1</version>
+      <version>3.2.2</version>
```
michael-o: This should be a seperate issue if not necesary for this PR.

pom.xml
```diff
@@ -185,15 +180,33 @@ under the License.
...
+        <scope>test</scope>
       </dependency>
     </dependencies>
   </dependencyManagement>
   <dependencies>
     <dependency>
       <groupId>junit</groupId>
       <artifactId>junit</artifactId>
-      <version>4.11</version>
+      <version>4.12</version>
```
michael-o: Same as commons-collections

doxia-modules/doxia-module-fo/src/main/java/org/apache/maven/doxia/module/fo/FoAggregateSink.java
```diff
@@ -1322,12 +1322,12 @@ private SinkEventAttributeSet getGraphicsAttributes( String logo )
         }
         catch ( IOException e )
         {
-            getLog().debug( e );
+            getLog().debug( e.getMessage() );
```
michael-o: Leave it as `e`, SFL4J and backends will do the rest for us.

michael-o:
@hboutemy Looks good to me. Tests pending...

hboutemy:
I'm not in favor of removing doxia-logging: deprecating and modifying it to use slf4j would be more compatible
I know there are not so much Doxia extensions in the wild: there is at least asciidoc I know of. I didn't check, but would be useful to not break this one at least

michael-o:
Why not deprecate in 1.8 and remove with 2.0?

hboutemy:
+1

michael-o:
@slachiewicz Can you spawn another ticket and create a new PR? We will leave this one open.

slachiewicz:
There is almost no usage of logging in Doxia, so any extention needs only to remove one call.

doxia-modules/doxia-module-fo/src/main/java/org/apache/maven/doxia/module/fo/FoAggregateSink.java
slachiewicz: there is no constructor with Exception, i will change to e.getMessage(), e

pom.xml
slachiewicz: ok, will be diffret pull - only with dependences

doxia-modules/doxia-module-fo/src/main/java/org/apache/maven/doxia/module/fo/FoAggregateSink.java
hazendaz: @slachiewicz @michael-o Proper code here would be ```getLog().debug("", e)```.  That assumes you want the full stack trace and not just the message.

slachiewicz:
New version with deprecation in #8

abelsromero:
> I'm not in favor of removing doxia-logging: deprecating and modifying it to use slf4j would be more compatible
> I know there are not so much Doxia extensions in the wild: there is at least asciidoc I know of. I didn't check, but would be useful to not break this one at least

I mantain the asciidoctor extension and I am fine with removing logging-api. In fact we have problems sharing code between the our normal Mojo and the doxia parser because one uses "maven.plugin.logging.Log" and the other "maven.doxia.logging.Log". Moving doxia to Sl4j won't fix it 100%, but will be one step closer.

PR #7, opened 2018-01-14
[DOXIA-409] Upgrade to FOP 2.2

slachiewicz:
Fix by: Michael Simacek <msimacek@redhat.com

Source: http://pkgs.fedoraproject.org/cgit/rpms/maven-doxia.git/tree/0004-Port-to-fop-2.0.patch

### DIFF
--- a/doxia-modules/doxia-module-fo/pom.xml
+++ b/doxia-modules/doxia-module-fo/pom.xml
@@ -85,7 +85,7 @@ under the License.
     <dependency>
       <groupId>org.apache.xmlgraphics</groupId>
       <artifactId>fop</artifactId>
-      <version>0.95</version>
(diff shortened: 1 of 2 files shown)


hboutemy:
interesting: did you try the result of this change with maven-pdf-plugin or other tools/plugins that produce pdf using FO?

slachiewicz:
Yes, i run tests inside doxia (some minor warnings) and today also tests with maven-pdf-plugin and all looks good.

hboutemy:
ok, let's have a try
I merged this PR and will update maven-pdf-plugin to use the new SNAPSHOT
then please ask on the dev mailing list for people to test maven-pdf-plugin or any other tool using this component to check that everything works well: we need to engage as many testers as possible

PR #11, opened 2018-06-07
Propagate macro related exceptions to caller in xhtml parser

ghilainm:
The exceptions are swallowed this is really unpractical when using the library (example building a site using mvn-site plugin).

### DIFF
--- a/doxia-modules/doxia-module-xhtml/src/main/java/org/apache/maven/doxia/module/xhtml/XhtmlParser.java
+++ b/doxia-modules/doxia-module-xhtml/src/main/java/org/apache/maven/doxia/module/xhtml/XhtmlParser.java
@@ -269,11 +269,11 @@ public class XhtmlParser
         }
         catch ( MacroExecutionException e )
         {
-            throw new XmlPullParserException( "Unable to execute macro in the document: " + macroName );
+            throw new XmlPullParserException( "Unable to execute macro in the document: " + macroName , e);
         }
         catch ( MacroNotFoundException me )
         {
-            throw new XmlPullParserException( "Macro not found: " + macroName );
+            throw new XmlPullParserException( "Macro not found: " + macroName, me);
         }
     }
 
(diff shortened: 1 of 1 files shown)


comtention:
I have exactly the same issue, it is really blocking for me.

michael-o:
A JIRA issue please and I will merge.

doxia-modules/doxia-module-xhtml/src/main/java/org/apache/maven/doxia/module/xhtml/XhtmlParser.java
```diff
@@ -269,11 +269,11 @@ private void processMacro( String text, Sink sink )
         }
         catch ( MacroExecutionException e )
         {
-            throw new XmlPullParserException( "Unable to execute macro in the document: " + macroName );
+            throw new XmlPullParserException( "Unable to execute macro in the document: " + macroName , null, e );
```
hboutemy: can you provide an example of failing macro for me to test in real conditions how the failure renders without and with the update, please?

hboutemy:
FYI, Jira issue created: https://issues.apache.org/jira/browse/DOXIA-585

doxia-modules/doxia-module-xhtml/src/main/java/org/apache/maven/doxia/module/xhtml/XhtmlParser.java
ghilainm: @hboutemy  See my comment above :)

doxia-modules/doxia-module-xhtml/src/main/java/org/apache/maven/doxia/module/xhtml/XhtmlParser.java
```diff
@@ -269,11 +269,11 @@ private void processMacro( String text, Sink sink )
...
         }
         catch ( MacroNotFoundException me )
         {
-            throw new XmlPullParserException( "Macro not found: " + macroName );
+            throw new XmlPullParserException( "Macro not found: " + macroName, null, me );
```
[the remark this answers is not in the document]
hboutemy: > It is also for me anyway a bad practice to swallow exceptions
unless you know why: let's not make it a pure dogma
for macro not found, it'll be easy to reproduce, I'll do it myself
that's more for execution exception that I don't know yet how to easily cause a failure: I'll find

thank you for the report, for sure, it has to be improved

hboutemy:
fixed in d977f4e4f4a9f7d5549f25daa56d3cd0f756023a
this not only keeps the exception in case of macro execution failure but also keep parser in every case to keep a link with the original document location

PR #17, opened 2019-05-28
[DOXIA-583] Avoid over-escaping table captions

harawata:
Please see https://issues.apache.org/jira/browse/DOXIA-583 for details.

### DIFF
--- a/doxia-core/src/main/java/org/apache/maven/doxia/sink/impl/Xhtml5BaseSink.java
+++ b/doxia-core/src/main/java/org/apache/maven/doxia/sink/impl/Xhtml5BaseSink.java
@@ -2434,7 +2434,7 @@ public class Xhtml5BaseSink
     {
         if ( !this.tableCaptionXMLWriterStack.isEmpty() && this.tableCaptionXMLWriterStack.getLast() != null )
         {
-            this.tableCaptionXMLWriterStack.getLast().writeText( unifyEOLs( text ) );
+            this.tableCaptionXMLWriterStack.getLast().writeMarkup( unifyEOLs( text ) );
         }
         else if ( !this.tableContentWriterStack.isEmpty() && this.tableContentWriterStack.getLast() != null )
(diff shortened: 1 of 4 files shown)


harawata:
There seems to be the same issue in Xhtml5Sink indeed.
I have already committed the fix.

By the way, I noticed that the existing assertions don't work as expected.

https://github.com/apache/maven-doxia/blob/e005c348ddd4367de8871f2e38d698b615db7dce/doxia-core/src/test/java/org/apache/maven/doxia/sink/impl/Xhtml5BaseSinkTest.java#L175-L186

They do `indexOf(...) != 1`, but the value should be `-1`, I think.

slachiewicz:
That's exactly what I had in mind. Your changes are consistent with my local changes.

Good catch with `indexOf` - we can replace it with just `contains` here in xhtml5 but also XHtmlBaseSink.
Could you also fix this issues and squash commits?

harawata:
Sure! =)
However, simply replacing `!= 1` with `contains` made some (?) assertions fail, so I'll take a deeper look at the literal parts tomorrow night.
Thanks for the review, @slachiewicz !

harawata:
Hi again,

As I mentioned, the 'expected value' of the assertions had to be modified to pass the tests after replacing `indexOf(...) != 1` with `contains()`.
Does the commit c8e834b649f7f7d5ee2b51c0f210c20c8e9bebdd look OK to you?
Or could it be related to [DOXIA-588](https://issues.apache.org/jira/projects/DOXIA/issues/DOXIA-588) maybe?

If it looks OK to you, please let me know and I'll do the squash & force-push.

slachiewicz:
looks good to me. Please squash commits related to table caption to one and separate only related to .contains. Thx for time spent on verification!

harawata:
Updated. Please let me know if there is anything you want me to change.
Thanks again for your time! =)

PR #19, opened 2020-03-01
[DOXIA-607] Add support for Xhtml5 keygen tag

minfrin:
This will allow maven generated documentation to include a demo of the keygen tag, as with other form elements.

### DIFF
--- a/doxia-core/src/main/java/org/apache/maven/doxia/markup/HtmlMarkup.java
+++ b/doxia-core/src/main/java/org/apache/maven/doxia/markup/HtmlMarkup.java
@@ -473,6 +473,17 @@ public interface HtmlMarkup
     /** Xhtml tag for <code>kbd</code>. */
     Tag KBD = Tag.KBD;
 
+    /** Html5 tag for <code>keygen</code>. */
+    Tag KEYGEN = new Tag()
+    {
+        /** {@inheritDoc} */
+        @Override
+        public String toString()
+        {
+            return "keygen";
+        }
+    };
+
     /** Xhtml tag for <code>label</code>. */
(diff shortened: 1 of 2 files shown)


michael-o:
Mozilla says this is obsolete: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/keygen

minfrin:
The tag is a formal part of the html5 spec as follows: https://www.w3.org/TR/2014/REC-html5-20141028/single-page.html#the-keygen-element

The need to both document and provide examples of the tag in use is in an attempt to bring the tag up to date with respect to current crypto standards. Not being able to include the  tag in doxia generated documentation is making our lives hard.

doxia-core/src/main/java/org/apache/maven/doxia/markup/HtmlMarkup.java
```diff
@@ -473,6 +473,17 @@ public String toString()
     /** Xhtml tag for <code>kbd</code>. */
     Tag KBD = Tag.KBD;
 
+    /** Html5 tag for <code>keygen</code>. */
```
elharo: Html5 --> HTML 5

michael-o: It is actually HTML5. See https://html.spec.whatwg.org/multipage/introduction.html#is-this-html5?

michael-o:
Surprinsingly, the most recent version does not mention a deprecation: https://html.spec.whatwg.org/#keygen

michael-o:
Branch pushed, awaiting tests...

jira-importer:
Resolve #572

PR #35, opened 2020-06-28
[DOXIA-614] support source reference in doxia parser

abelsromero:
There's still a thing to fix with test begin discussed in https://issues.apache.org/jira/browse/DOXIA-614.
Other than that, the changes are all done and the tests pass. I tried to minimize changes so no tests have been modified and all works as-is.
To facilitate review, here is a diagram with changes. Green for added elements, red for removed.
doxia-modules-classes.

Some observations open for discussion:
- Used null insteall of empty string to represent there's no reference at all. I found the empty string to be ambiguous. I noted that `AptParser`, `ConfluenceParser` and `TwikiParser` use empty string, but I consider this implementation specific.

### DIFF
--- a/doxia-core/src/main/java/org/apache/maven/doxia/Doxia.java
+++ b/doxia-core/src/main/java/org/apache/maven/doxia/Doxia.java
@@ -68,7 +68,7 @@ public interface Doxia
      * Return a parser for the given <code>parserId</code>.
      *
      * @param parserId identifier for the parser to use
-     * @return the parser defining by parserId
+     * @return the parser identified by parserId
      * @throws ParserNotFoundException if no parser could be found for the given id
      */
     Parser getParser( String parserId )
--- a/doxia-core/src/main/java/org/apache/maven/doxia/Doxia.java
+++ b/doxia-core/src/main/java/org/apache/maven/doxia/Doxia.java
@@ -41,13 +41,11 @@ public interface Doxia
      * Parses the given source model using a parser with given id,
      * and emits Doxia events into the given sink.
      *
-     * @param source not null reader that provides the source document.
-     * You could use <code>newReader</code> methods from {@link org.codehaus.plexus.util.ReaderFactory}.
-     * @param parserId Identifier for the parser to use.
-     * @param sink A sink that consumes the Doxia events.
-     * @throws org.apache.maven.doxia.parser.manager.ParserNotFoundException
-     * if no parser could be found for the given id.
-     * @throws org.apache.maven.doxia.parser.ParseException if the model could not be parsed.
+     * @param source not null reader that provides the source document
+     * @param parserId identifier for the parser to use
+     * @param sink a sink that consumes the Doxia events
+     * @throws ParserNotFoundException if no parser could be found for the given id
+     * @throws ParseException if the model could not be parsed
      */
     void parse( Reader source, String parserId, Sink sink )
         throws ParserNotFoundException, ParseException;
@@ -56,14 +54,12 @@ public interface Doxia
      * Parses the given source model using a parser with given id,
      * and emits Doxia events into the given sink.
      *
-     * @param source not null reader that provides the source document.
-     * You could use <code>newReader</code> methods from {@link org.codehaus.plexus.util.ReaderFactory}.
-     * @param parserId Identifier for the parser to use.
-     * @param sink A sink that consumes the Doxia events.
-     * @param reference A string containing the reference to the source of the input string (e.g. filename).
-     * @throws org.apache.maven.doxia.parser.manager.ParserNotFoundException
-     * if no parser could be found for the given id.
-     * @throws org.apache.maven.doxia.parser.ParseException if the model could not be parsed.
+     * @param source not null reader that provides the source document
+     * @param parserId identifier for the parser to use
+     * @param sink a sink that consumes the Doxia events
+     * @param reference string containing the reference to the source (e.g. filename)
+     * @throws ParserNotFoundException if no parser could be found for the given id
+     * @throws ParseException if the model could not be parsed
      */
     void parse( Reader source, String parserId, Sink sink, String reference )
         throws ParserNotFoundException, ParseException;
@@ -71,10 +67,9 @@ public interface Doxia
     /**
      * Return a parser for the given <code>parserId</code>.
      *
-     * @param parserId Identifier for the parser to use.
-     * @return the parser defining by parserId.
-     * @throws org.apache.maven.doxia.parser.manager.ParserNotFoundException
-     * if no parser could be found for the given id.
+     * @param parserId identifier for the parser to use
+     * @return the parser defining by parserId
+     * @throws ParserNotFoundException if no parser could be found for the given id
      */
     Parser getParser( String parserId )
         throws ParserNotFoundException;
--- a/doxia-core/src/main/java/org/apache/maven/doxia/Doxia.java
+++ b/doxia-core/src/main/java/org/apache/maven/doxia/Doxia.java
@@ -53,6 +53,22 @@ public interface Doxia
         throws ParserNotFoundException, ParseException;
 
     /**
+     * Parses the given source model using a parser with given id,
+     * and emits Doxia events into the given sink.
+     *
+     * @param source not null reader that provides the source document.
+     * You could use <code>newReader</code> methods from {@link org.codehaus.plexus.util.ReaderFactory}.
+     * @param parserId Identifier for the parser to use.
+     * @param sink A sink that consumes the Doxia events.
+     * @param reference A string containing the reference to the source of the input string (e.g. filename).
+     * @throws org.apache.maven.doxia.parser.manager.ParserNotFoundException
+     * if no parser could be found for the given id.
+     * @throws org.apache.maven.doxia.parser.ParseException if the model could not be parsed.
+     */
+    void parse( Reader source, String parserId, Sink sink, String reference )
+        throws ParserNotFoundException, ParseException;
+
+    /**
      * Return a parser for the given <code>parserId</code>.
      *
      * @param parserId Identifier for the parser to use.
--- a/doxia-core/src/main/java/org/apache/maven/doxia/parser/AbstractParser.java
+++ b/doxia-core/src/main/java/org/apache/maven/doxia/parser/AbstractParser.java
@@ -101,7 +101,7 @@ public abstract class AbstractParser
     /**
      * {@inheritDoc}
      *
-     * @return a int.
+     * @return a int
      */
     public int getType()
     {
@@ -117,7 +117,7 @@ public abstract class AbstractParser
     /**
      * <p>isEmitComments.</p>
      *
-     * @return a boolean.
+     * @return a boolean
      */
     public boolean isEmitComments()
     {
@@ -127,11 +127,11 @@ public abstract class AbstractParser
     /**
      * Execute a macro on the given sink.
      *
-     * @param macroId An id to lookup the macro.
-     * @param request The corresponding MacroRequest.
-     * @param sink The sink to receive the events.
-     * @throws org.apache.maven.doxia.macro.MacroExecutionException if an error occurred during execution.
-     * @throws org.apache.maven.doxia.macro.manager.MacroNotFoundException if the macro could not be found.
+     * @param macroId an id to lookup the macro
+     * @param request the corresponding MacroRequest
+     * @param sink the sink to receive the events
+     * @throws org.apache.maven.doxia.macro.MacroExecutionException if an error occurred during execution
+     * @throws org.apache.maven.doxia.macro.manager.MacroNotFoundException if the macro could not be found
      */
     // Made public right now because of the structure of the APT parser and
     // all its inner classes.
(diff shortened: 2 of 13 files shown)


doxia-core/src/main/java/org/apache/maven/doxia/Doxia.java
```diff
@@ -52,6 +52,22 @@
     void parse( Reader source, String parserId, Sink sink )
         throws ParserNotFoundException, ParseException;
 
+    /**
+     * Parses the given source model using a parser with given id,
+     * and emits Doxia events into the given sink.
+     *
+     * @param source not null reader that provides the source document.
+     * You could use <code>newReader</code> methods from {@link org.codehaus.plexus.util.ReaderFactory}.
```
elharo, line 60: I'd delete this note. I'd prefer not to encourage extra plexus dependencies

doxia-core/src/main/java/org/apache/maven/doxia/Doxia.java
```diff
@@ -52,6 +52,22 @@
...
+     * @param parserId Identifier for the parser to use.
```
elharo, line 61: nit: no caps and no period, per Oracle guidelines

doxia-core/src/main/java/org/apache/maven/doxia/Doxia.java
```diff
@@ -52,6 +52,22 @@
...
+     * @param sink A sink that consumes the Doxia events.
```
elharo, line 62: ditto

doxia-core/src/test/java/org/apache/maven/doxia/DefaultDoxiaTest.java
```diff
@@ -0,0 +1,54 @@
...
+
+    @Test
+    public void testFailsWhenParserIdDoesNotExist() throws Exception
+    {
+        final String parserId = "a-parser";
+        final Doxia doxia = lookup( Doxia.class );
+
+        try
+        {
+            doxia.getParser( parserId );
```
elharo, line 45: test should fail here if exception is not thrown

doxia-core/src/test/java/org/apache/maven/doxia/DefaultDoxiaTest.java
```diff
@@ -0,0 +1,54 @@
...
+        }
+        catch ( Exception e )
```
elharo, line 47: just catch ParserNotFoundException rather than asserting the type

doxia-core/src/main/java/org/apache/maven/doxia/Doxia.java
abelsromero: The exceptions in the oracle reference (https://www.oracle.com/technical-resources/articles/java/javadoc-tool.html#@exception) do capitalize, should I correct exceptions also accordingly?

abelsromero:
Applied the suggestions tried to keep commit history clean.
- Squashed test change in Test commit
- Added new commit for Doxia's JavaDoc review

doxia-core/src/main/java/org/apache/maven/doxia/parser/AbstractParser.java
```diff
@@ -179,15 +178,32 @@ protected File getBasedir()
...
     }
-    
+
+    /**
+     * {@inheritDoc}
+     *
+     * Convenience method to parse an arbitrary string and emit events into the given sink.
+     *
+     * @param string A string that provides the source input.
+     * @param sink A sink that consumes the Doxia events.
```
elharo, line 190: A --> a
no period

ditto below

doxia-core/src/main/java/org/apache/maven/doxia/parser/AbstractXmlParser.java
```diff
@@ -100,7 +98,7 @@
     private boolean validate = false;
 
     /** {@inheritDoc} */
-    public void parse( Reader source, Sink sink )
+    public void parse( Reader source, Sink sink, String reference )
```
elharo, line 101: Does this change the public API? I'm not sure these changes are compatible.

abelsromero: Please, refer to the JIRA for the conversation https://issues.apache.org/jira/browse/DOXIA-614

doxia-core/src/main/java/org/apache/maven/doxia/parser/AbstractParser.java
abelsromero: If all the JavaDocs need to be reviewed, these maybe should be addressed in a specific effort with a specific JIRA issue. I just added the information related to the new parameter `reference` following the patterns already pressent in the project.

doxia-core/src/main/java/org/apache/maven/doxia/Doxia.java
```diff
@@ -41,24 +41,35 @@
...
+    /**
+     * Parses the given source model using a parser with given id,
+     * and emits Doxia events into the given sink.
+     *
+     * @param source not null reader that provides the source document
+     * @param parserId identifier for the parser to use
+     * @param sink a sink that consumes the Doxia events
+     * @param reference string containing the reference to the source (e.g. filename)
+     * @throws ParserNotFoundException If no parser could be found for the given id
+     * @throws ParseException If the model could not be parsed
```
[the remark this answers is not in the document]
abelsromero: Please refer to my previous question https://github.com/apache/maven-doxia/pull/35#discussion_r446685351

doxia-core/src/main/java/org/apache/maven/doxia/Doxia.java
elharo: I assume that's an oversight since it contradicts the language of the doc and only appears in an example

doxia-core/src/main/java/org/apache/maven/doxia/parser/AbstractParser.java
```diff
@@ -37,6 +29,13 @@
 import org.apache.maven.doxia.sink.Sink;
 import org.codehaus.plexus.component.annotations.Requirement;
 
+import java.io.File;
+import java.io.IOException;
+import java.io.InputStream;
+import java.io.Reader;
+import java.io.StringReader;
+import java.util.Properties;
+
```
hboutemy, line 38: is it possible to avoid such reformatting of imports, please?

doxia-core/src/main/java/org/apache/maven/doxia/parser/AbstractXmlParser.java
```diff
@@ -49,18 +32,33 @@
...
 import org.codehaus.plexus.util.StringUtils;
 import org.codehaus.plexus.util.xml.pull.MXParser;
 import org.codehaus.plexus.util.xml.pull.XmlPullParser;
 import org.codehaus.plexus.util.xml.pull.XmlPullParserException;
-
 import org.xml.sax.EntityResolver;
 import org.xml.sax.InputSource;
 import org.xml.sax.SAXException;
 
+import java.io.BufferedReader;
```
hboutemy, line 45: is it possible to avoid such reformatting of imports, please?

doxia-core/src/main/java/org/apache/maven/doxia/parser/AbstractParser.java
abelsalgadoromeroWD: It's by Intellij, I'll restore it in a commit. I have noted many other things that even with the style configured as described in https://maven.apache.org/developers/conventions/code.html#intellij-idea. For example exception below the method definition are indented at 8 spaces while the original code is at 4.
I try to keep some things that I see, but some may slipt, sorry.

abelsromero: It's by Intellij, I'll restore it in a commit. I have noted many other things that even with the style configured as described in https://maven.apache.org/developers/conventions/code.html#intellij-idea. For example exception below the method definition are indented at 8 spaces while the original code is at 4.
I try to keep some things that I see, but some may slipt, sorry.

abelsromero:
All improvements applied, thanks for your time.
Note there are a couple of points for discussion I added in the JIRA issue. I am new, so let me know where do you think is best to discuss, directly here or there.

abelsromero:
Applied the suggestions. Anything else that could be improved?

doxia-core/src/main/java/org/apache/maven/doxia/parser/AbstractXmlParser.java
elharo: The reference argument doesn't seem to be used here.

abelsromero: Effectively, it is not. As explained in the JIRA issue the PR aims to be that the default method instead of the other one with 2 arguments. The way it is implemented right now, even when `Parser` interface has two method, the one with thre is not usable by how it's overriden in `AbstractParser` and never invoked in `DefaultDoxia`.
The final goal is to deprecate the one with two arguments so that any plugin can obtain the metadata of the sourcefile bien processed.

I am available to discuss online if anyone needs to, because I am bit confused by how this review is proceding. I think the JIRA issue, explanation here and the additional diagram provide a clear description of the goals and changes done which are the complex topics. However, I am seeing a focus on minor things. Furthermore, I raised 2 important points in the JIRA issue which had had no answer and which imo are worth discussing, one related to problems with the build and the other with testing.

slachiewicz:
ok, now we have green build (clirr update was missing) 
CI: https://ci-builds.apache.org/job/Maven/job/maven-box/job/maven-doxia/job/DOXIA-614%252Fsupport-source-reference-in-doxia-parser/
branch: https://github.com/apache/maven-doxia/commits/DOXIA-614/support-source-reference-in-doxia-parser

I'm ok with changes, but it would b good to bump to 1.10 with proposed change  wdyt @hboutemy ?

doxia-core/src/main/java/org/apache/maven/doxia/parser/Xhtml5BaseParser.java
```diff
@@ -81,14 +81,14 @@
 
     /** {@inheritDoc} */
     @Override
-    public void parse( Reader source, Sink sink )
+    public void parse( Reader source, Sink sink, String reference )
```
elharo, line 84: It's hard to follow from the diff, but are any of the apparent method signature changes real? Or is it just Github diff being funky?

If these are real changes is it possible to add overloads with the old variants that pass null for reference?

abelsromero: Not really.

To see the changes probably, the image in the PR description summarizes the changes (@ means overrride method), at least it helped me to track the changes.
The problems is the `parse(String,String,String)` method in `AbstractParser` was calling `parse(String, String)` without the third argument. So the "reference" was totally lost and no one could use it.
Even with fixing the call in `AbstractParser`, the problem is that all implementations add logic in their respective `parse` and relly on the others using `super.parse(String,String)` calls. So only that change would mean that someone overriding `Xhtml5Parser::parse(String,String,String)` would find a total different behauviour than extending the 2 args method because the will be extending the `AbstractParser` really and missing a lot of logic.

The only solution I saw was to replace 2 args method by the 3 args in all implementations and call the 3 args in the `AbstractParser` with null. That makes `reference` available and should not brake current doxia components.

abelsromero:
Anything I should review?

elharo:
If you resolve conflicts and ping me, I can take another look.

abelsromero:
Done :+1:

doxia-core/src/main/java/org/apache/maven/doxia/Doxia.java
```diff
@@ -41,24 +41,35 @@
...
+     * and emits Doxia events into the given sink.
+     *
+     * @param source not null reader that provides the source document
+     * @param parserId identifier for the parser to use
+     * @param sink a sink that consumes the Doxia events
+     * @param reference string containing the reference to the source (e.g. filename)
+     * @throws ParserNotFoundException if no parser could be found for the given id
+     * @throws ParseException if the model could not be parsed
+     */
+    void parse( Reader source, String parserId, Sink sink, String reference )
```
elharo, line 64: probably doesn't matter a great deal, but adding methods to interfaces is not backwards compatible

doxia-core/src/main/java/org/apache/maven/doxia/parser/AbstractParser.java
```diff
@@ -101,7 +101,7 @@
     /**
      * {@inheritDoc}
      *
-     * @return a int.
+     * @return a int
```
elharo, line 104: I'd just delete this, if there's nothing more to say than the type

doxia-core/src/main/java/org/apache/maven/doxia/parser/AbstractParser.java
```diff
@@ -117,7 +117,7 @@ public void setEmitComments( boolean emitComments )
     /**
      * <p>isEmitComments.</p>
      *
-     * @return a boolean.
+     * @return a boolean
```
elharo, line 120: ditto

elharo (approved):

elharo (changes requested):
Jenkins failed: https://ci-builds.apache.org/job/Maven/job/maven-box/job/maven-doxia/job/DOXIA-614B/

slachiewicz:
ok, fixed the issue and made a small adjustment. Build gene https://ci-builds.apache.org/job/Maven/job/maven-box/job/maven-doxia/job/DOXIA-614%252Fsupport-source-reference-in-doxia-parser/

PR #50, opened 2021-01-03
[DOXIA-542] Quotes vs apostrophes

bertysentry:
This PR solves [DOXIA-542](https://issues.apache.org/jira/browse/DOXIA-542) by disabling the replacement of "single" apostrophes with stylized quotes.

`This isn't a quote` will keep the good old apostrophe.
`This is 'a quote'` will change the apostrophes to opening and closing quotes.


jira-importer:
Resolve #616

PR #53, opened 2021-01-07
rollback Java 7 breaking change

elharo:
@michael-o @bertysentry


bertysentry:
Personally, I'm in favor of simply adding a commit that fixes the problem rather than reverting and then recommitting the same code plus the fix. But that's a git philosophy question, so I'll let you guys decide ;-)

elharo:
That's a moot point until such a fix exists.

bertysentry:
I have it. I'll submit my PR in a moment.

michael-o:
Yepp, I need also a day or two to digest.

bertysentry:
No problem. Remember we also have [this PR](https://github.com/apache/maven-doxia/pull/54) that restores the Java 7 compatibility. Thanks!

PR #56, opened 2021-02-06
Maven 3.1.1 Prepare switch plexus-container-default to org.eclipse.si…

slachiewicz:
…su.plexus


slachiewicz:
initial commit to fix known differences with Sisu Plexus. 

Final commit will change org.codehaus.plexus:plexus-container-default to org.eclipse.sisu:org.eclipse.sisu.plexus
In all Maven libs/plugins - from Maven 3.1 (ie switch to Eclipse Sisu) we are no longer use plexus-container-default

michael-o:
Very very nice...

PR #61, opened 2021-04-26
[DOXIA-588] The Sink.JUSTIFY_XXX was ignored

qualitesys:
See Jira ticket https://issues.apache.org/jira/browse/DOXIA-588
Source code patch
Update Unit test as well

### DIFF
--- a/.github/workflows/maven.yml
+++ b/.github/workflows/maven.yml
@@ -1,6 +1,21 @@
-# This workflow will build a Java project with Maven
-# For more information see: https://help.github.com/actions/language-and-framework-guides/building-and-testing-java-with-maven
-
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
 name: Java CI with Maven
 
 on:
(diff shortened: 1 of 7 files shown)


.github/workflows/maven.yml
```diff
@@ -0,0 +1,44 @@
+ # Licensed to the Apache Software Foundation (ASF) under one
```
elharo, line 1: This should be in a separate PR as it's not directly related to fixing this bug

doxia-core/src/main/java/org/apache/maven/doxia/sink/impl/Xhtml5BaseSink.java
```diff
@@ -1658,6 +1659,35 @@ public void tableHeaderCell( SinkEventAttributes attributes )
     private void tableCell( boolean headerRow, MutableAttributeSet attributes )
     {
         Tag t = ( headerRow ? HtmlMarkup.TH : HtmlMarkup.TD );
+        
+        if (   !headerRow 
+            && null != this.cellCountStack
```
elharo, line 1664: this.foo != null
is far more common and unsurprising. Putting the literal first is a C convention not needed in Java.

elharo (commented):
running through Jenkins: https://ci-builds.apache.org/job/Maven/job/maven-box/job/maven-doxia/job/DOXIA-588/

elharo (changes requested):
Jenkins passed but .github/workflows/maven.yml needs to be moved to a separate PR

.github/workflows/maven.yml
qualitesys: I am not a GitHub expert at all. I can remove maven.yml in my repo, no problem. Is it enough for this PR ?

doxia-core/src/main/java/org/apache/maven/doxia/sink/impl/Xhtml5BaseSink.java
qualitesys: Starting with constants in logical expressions prevents from a very common mistakes that can be very tricky, this concerns Java, C, C++, C# att least.

See [https://cwe.mitre.org/data/definitions/481.html](url)

elharo: In C, yes, Not in Java because == is used for comparison instead of =. The only way this can occur in Java is if you're comparing to a boolean and in that case you should just use the boolean directly.

.github/workflows/maven.yml
elharo: you need to push too.

slachiewicz:
thx for the patch, I merged it with minor changes. 
I also added changes to Github Actions based on our other projects 794c39dbee17a35e2e74bf001450b2c3cf15959a

PR #65, opened 2021-05-30
replace deprecated class from commons-lang

elharo:


michael-o (approved):

slachiewicz (approved):

elharo:
Build windows-jdk11 looks borked again :-(

doxia-core/pom.xml
```diff
@@ -59,6 +59,11 @@ under the License.
       <artifactId>commons-lang3</artifactId>
       <version>3.8.1</version>
     </dependency>
+    <dependency>
+      <groupId>org.apache.commons</groupId>
+      <artifactId>commons-text</artifactId>
+      <version>1.3</version>
```
elilja, line 65: Why such an old version of commons-text?

elharo: Because everything newer requires Java 8. :-(

elilja: That's fine since some time ago, I believe. You may require Java 8 in plugins and components etc for the Maven eco system

PR #98, opened 2022-05-18
[DOXIA-590] Either provided element class or default class gets ignored

michael-o:
https://issues.apache.org/jira/browse/DOXIA-590
Incomplete:
* `onSection()` is missing
* Additional tests are missing

### DIFF
--- a/doxia-core/src/main/java/org/apache/maven/doxia/sink/impl/Xhtml5BaseSink.java
+++ b/doxia-core/src/main/java/org/apache/maven/doxia/sink/impl/Xhtml5BaseSink.java
@@ -1567,21 +1567,24 @@ public class Xhtml5BaseSink
     @Override
     public void tableRow( SinkEventAttributes attributes )
     {
-        MutableAttributeSet att = new SinkEventAttributeSet();
+        MutableAttributeSet atts = SinkUtils.filterAttributes(
+                attributes, SinkUtils.SINK_TR_ATTRIBUTES );
 
-        if ( evenTableRow )
+        if ( atts == null )
         {
-            att.addAttribute( Attribute.CLASS, "a" );
-        }
-        else
-        {
-            att.addAttribute( Attribute.CLASS, "b" );
+            atts = new SinkEventAttributeSet();
         }
 
-        att.addAttributes( SinkUtils.filterAttributes(
-                attributes, SinkUtils.SINK_TR_ATTRIBUTES  ) );
+        String rowClass = evenTableRow ? "a" : "b";
+        if ( atts.isDefined( Attribute.CLASS.toString() ) )
+        {
+            String givenRowClass = (String) atts.getAttribute( Attribute.CLASS.toString() );
+            rowClass = givenRowClass + " " + rowClass;
+        }
 
-        writeStartTag( HtmlMarkup.TR, att );
+        atts.addAttribute( Attribute.CLASS, rowClass );
+
+        writeStartTag( HtmlMarkup.TR, atts );
 
         evenTableRow = !evenTableRow;
 
@@ -1833,44 +1836,28 @@ public class Xhtml5BaseSink
         }
     }
 
-    /** {@inheritDoc} */
+    /**
+     * The default style class for external link is <code>externalLink</code>.
+     *
+     * {@inheritDoc}
+     * @see javax.swing.text.html.HTML.Tag#A
+     **/
     @Override
     public void link( String name )
     {
         link( name, null );
     }
 
-    /** {@inheritDoc} */
+    /**
+     * The default style class for external link is <code>externalLink</code>.
+     *
+     * {@inheritDoc}
+     * @see javax.swing.text.html.HTML.Tag#A
+     **/
     @Override
     public void link( String name, SinkEventAttributes attributes )
     {
-        if ( attributes == null )
-        {
-            link( name, null, null );
-        }
-        else
-        {
-            String target = (String) attributes.getAttribute( Attribute.TARGET.toString() );
-            MutableAttributeSet atts = SinkUtils.filterAttributes(
-                    attributes, SinkUtils.SINK_LINK_ATTRIBUTES  );
-
-            link( name, target, atts );
-        }
-    }
-
-    /**
-     * Adds a link with an optional target.
-     * The default style class for external link is <code>externalLink</code>.
-     *
-     * @param href the link href.
-     * @param target the link target, may be null.
-     * @param attributes an AttributeSet, may be null.
-     *      This is supposed to be filtered already.
-     * @see javax.swing.text.html.HTML.Tag#A
-     */
-    private void link( String href, String target, MutableAttributeSet attributes )
-    {
-        if ( href == null )
+        if ( name == null )
         {
             throw new NullPointerException( "Link name cannot be null!" );
         }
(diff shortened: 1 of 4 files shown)


feckertson:
The code changes cover what I was described in the Jira. 
There should probably be some new unit tests for the tableRow case.

doxia-core/src/main/java/org/apache/maven/doxia/sink/impl/Xhtml5BaseSink.java
```diff
@@ -1567,21 +1567,24 @@ public void tableRow()
...
+            atts = new SinkEventAttributeSet();
         }
-        else
+
+        String rowClass = evenTableRow ? "a" : "b";
+        if ( atts.isDefined( Attribute.CLASS.toString() ) )
         {
-            att.addAttribute( Attribute.CLASS, "b" );
+            String givenRowClass = (String) atts.getAttribute( Attribute.CLASS.toString() );
+            rowClass = givenRowClass + " " + rowClass;
```
feckertson, line 1582: This change covers the opening statement in my Jira but not the complexity described in my follow-up statement.

I believe the code should check to see if givenRowClass matches `"\b(?:a|b)\b"` and skip appending rowClass if it does. That is, don't add "a" or "b" if the provide class already has one or the other.  This would allow the consumer to decide whether a row is initially odd or even.

Why would one want to control this?  In order to create a table which supports hiding/showing selected rows and have it initially displayed with the rows hidden.

To be clear, the zebra striping does not work correctly when rows are hidden. To address the matter, javascript must be employed to fix up the "a" and "b" assignments when rows are hidden/displayed and I did not see a way to have doxia execute some javascript just before or just after the document is rendered. 

Without this control one must first display all the rows and then hide the rows that should be hidden.  This can create a noticeable flicker.  In my use case, I opted to start with no rows hidden because of the flicker.  The user then has to select a toggle to hide the noise, but it would be better experience to start with those rows already hidden.

michael-o: I see and understand your usecase. We have these options:
* Always append default class (the internal implementation might employ logic to determine the class name which is impossible to do externally)
* If external class attribute is provided, take it as-is, don't add default class
* Merge and avoid duplicates

For me personally, if you have decided to provide attributes which are being set in the method called you are making a deliberate decision to ignore the set values. One of the problems is that class is a merge attribute while others are simply replace. That makes it complex. WDYT?

@hboutemy Opinion?

feckertson: The second option is easy to comprehend and offers ultimate flexibility, **but** it forces a consumer to keep track of row parity just to put an additional decoration on a single row.... not very appealing.

The first option is what I originally proposed and what has been implemented, but it does not support a variety of use cases.  Zebra striping based on row number parity is not always correct. This is the fundamental dilemma.

The third option blends both but adds complexity and a performance hit.

Here is a potential fourth option:
Create a new tableRow method in SinkAdapter having additional parameters and implement the existing method to invoke it with reasonable defaults.  There are various choices for the extra parameter(s), but the following offers a lot of flexibility while always appending the "a" or "b": 
`tableRow(SinkEventAttributes attributes, boolean flipRowParity)`.


Use Cases:
- bold face just one row: pass true (or equivalently just call the original method).
- add two rows at a time planning for only one or the other to be visible at any one time: pass false for the first and true for the second.
- have multiple consecutive rows behave as a single row (same stripe color): pass true for the last in each group, and false for the rest.
- start off with some rows hidden and have a toggle to show/hide them: pass true for the displayed rows and false for the hidden ones.

`tableRow(SinkEventAttributes attributes, boolean appendStripeClass, boolean flipRowParity)` would be even more flexible, but the only situation I see where it makes sense to not add the "a" or "b" is if one want's to have more than two stripe colors.



I think all the other places which apply a default class can just append it (bodyTable, externalLink, section).

michael-o: The issue you depict isn't only about `tableRow()`  there are other methods as well. Therefore, I'd like to have a consistent solution for all of them instead of for `tableRow()` only.

feckertson: I am a fan of consistency and common code, but as far as I can tell the other methods do not use more than one "default class" so tableRow is very different from them.   Also, I do not believe it is a problem if a single class name is specified multiple times in an (X)HTML class string so the net effect is the same for merge and always append when the thing being appended never changes.

The only option which could produce a different result when the "default class" never changes is "omit if a class string is provided", but that just means "don't apply the default class style" which can be achieved with css.  

The first additional tableRow method I discussed is not really that different anyways.  It will always apply "a" or "b", The consumer just gets to control when it flips from one to the other. 

The alternate additional tableRow method does feel a little different, but it predicates on whether or not support for more than two stripe colors should be supported out-of-the-box. Has anyone ever asked for such a thing?  Besides, it could also be achieved with css: override "a" and "b" to do nothing then define and apply your own a', b', c'.

michael-o: Let me get back to you when I have sorted out issues in Maven SCM.

michael-o: Still waiting for @hboutemy's opinion.

hboutemy:
I don't understand anything, sorry: what is the issue? what does the code change (apart from order in attributes)?

michael-o:
> I don't understand anything, sorry: what is the issue? what does the code change (apart from order in attributes)?

One of the basic problems is that you cannot provide a CSS class attribute from outside because it is overridden so you cannot customize. It is not merged, but simply wiped out. This problem existed now for a very long time, I knew that, but this is the first issue regarding this.
Stupid simple: https://github.com/apache/maven-doxia/blob/c098cc9972b13c761091e66a57261e89195a9271/doxia-core/src/main/java/org/apache/maven/doxia/sink/impl/Xhtml5BaseSink.java#L1533-L1544

You pass the attribute set, but your potential, custom `CLASS` is dropped in any case.

hboutemy:
IIUC, currently BaseSink *replaces* the class attribute value with its own a or b, it should *append*?

michael-o:
> IIUC, currently BaseSink _replaces_ the class attribute value with its own a or b, it should _append_? I think a well named method could make sense

Yes, it blindly replaces instead taking care of the additive semantics of `class`. That is logically wrong.

michael-o:
@feckertson I'd like to settle this in the next couple of weeks. Please make a concrete proposal/PR.

feckertson:
@michael-o PR #104 improves this PR to allow the `hidden` class to decide whether the even/odd row parity gets toggled which is the gist of the matter.  If one wants ternary+ coloring, one can override the `a` and `b` css to do nothing and define/manage alternate classes based on visible row number mod N. 

Ideally org.apache.maven.skins:maven-skins should be improved to hide the 'hidden' rows, but one can provide additional css to do that in the meantime.

PR #102, opened 2022-05-28
[DOXIA-652] Drop build-info.properties in favor of default pom.proper…

michael-o:
…ties

Remove corresponding test because properties are only available in the
JAR after the tests have been run.

This closes #102


hboutemy (approved):
good simplification

PR #104, opened 2022-06-02
[DOXIA-590] implement hidden rows

feckertson:


feckertson:
Decides whether to toggle the even/odd table row parity for zebra striping based on whether the 'hidden' class is being applied.

Also adds unit tests for setting an externalLink and a class string simultaneously.

doxia-core/src/main/java/org/apache/maven/doxia/sink/impl/Xhtml5BaseSink.java
```diff
@@ -61,17 +61,20 @@
...
     /** An indication on if we're inside a head. */
     private boolean headFlag;
 
     /** Keep track of the main and div tags for content events. */
-    protected Stack<Tag> contentStack = new Stack<>();
+    protected Stack<Tag> contentStack = new Stack<Tag>();
 
     /** Keep track of the closing tags for inline events. */
-    protected Stack<List<Tag>> inlineStack = new Stack<>();
+    protected Stack<List<Tag>> inlineStack = new Stack<List<Tag>>();
```
michael-o, line 77: Why?

doxia-core/src/main/java/org/apache/maven/doxia/sink/impl/Xhtml5BaseSink.java
```diff
@@ -61,17 +61,20 @@
...
+
     /** Used to collect text events mainly for the head events. */
     private StringBuffer textBuffer = new StringBuffer();
 
     /** An indication on if we're inside a head. */
     private boolean headFlag;
 
     /** Keep track of the main and div tags for content events. */
-    protected Stack<Tag> contentStack = new Stack<>();
+    protected Stack<Tag> contentStack = new Stack<Tag>();
```
michael-o, line 74: Why?

doxia-core/src/main/java/org/apache/maven/doxia/sink/impl/Xhtml5BaseSink.java
```diff
@@ -136,13 +139,13 @@ public Xhtml5BaseSink( Writer out )
...
-        this.tableCaptionWriterStack = new LinkedList<>();
-        this.tableCaptionXMLWriterStack = new LinkedList<>();
-        this.tableCaptionStack = new LinkedList<>();
+        this.cellJustifStack = new LinkedList<int[]>();
+        this.isCellJustifStack = new LinkedList<Boolean>();
+        this.cellCountStack = new LinkedList<Integer>();
+        this.tableContentWriterStack = new LinkedList<StringWriter>();
+        this.tableCaptionWriterStack = new LinkedList<StringWriter>();
+        this.tableCaptionXMLWriterStack = new LinkedList<PrettyPrintXMLWriter>();
+        this.tableCaptionStack = new LinkedList<String>();
```
michael-o, line 148: Why?

michael-o (commented):
Why do I see the diamond changes?

michael-o:
Will check in the next couple of dazs.

feckertson:
Apologies for not checking the diff before creating the PR.  The apache format introduced unintended changes/noise. The standard format should be applied, but in a PR which only does that.

I couldn't find standard eclipse settings (code cleanup) for this org/project to avoid the diamond changes?  Are there any?

michael-o (changes requested):
I don't see handling for `onSection`.

doxia-core/src/test/java/org/apache/maven/doxia/sink/impl/Xhtml5BaseSinkTest.java
```diff
@@ -1086,15 +1086,86 @@ public void testTableRow()
...
+            sink.tableRow_();
+            sink.tableRows_();
+            sink.table_();
+        }
+        finally
+        {
+            sink.close();
+        }
+
+        String nl = System.lineSeparator();
```
michael-o, line 1102: Use provided `LS`

doxia-core/src/test/java/org/apache/maven/doxia/sink/impl/Xhtml5BaseSinkTest.java
```diff
@@ -1086,15 +1086,86 @@ public void testTableRow()
...
+            sink.tableRow();
+            sink.tableRow_();
         }
         finally
         {
             sink.close();
         }
 
-        assertEquals( "<tr style=\"bold\" class=\"a\"></tr>", writer.toString() );
+        String nl = System.lineSeparator();
```
michael-o, line 1154: Use provided `LS`

michael-o (changes requested):
A few nits. Otherwise your change looks reasonable. @hboutemy Any objections?

feckertson:
>  I don't see handling for onSection.

Please explain. I see no issues with the way onSection is coded (unless the concern is that onSection does nothing if the level is greater than 5).

michael-o:
> > I don't see handling for onSection.
> 
> Please explain. I see no issues with the way onSection is coded (unless the concern is that onSection does nothing if the level is greater than 5).

See here: https://github.com/apache/maven-doxia/blob/8ef6a26a35946385ad55193a63e0180a5d5320ec/doxia-core/src/main/java/org/apache/maven/doxia/sink/impl/XhtmlBaseSink.java#L472-L484

michael-o:
Merged. I have ignored `onSection()` because it will be removed in 2.0 anyway.

PR #106, opened 2022-07-07
[DOXIA-662] Generate unique ids

kwin:


doxia-core/src/main/java/org/apache/maven/doxia/index/IndexingSink.java
```diff
@@ -69,6 +72,10 @@
     /** The stack. */
     private final Stack<IndexEntry> stack;
 
+    /** A map containing all used ids of index entries as key and how often they are used as value
+     * (0-based, i.e. 0 means used 1 time). */
+    private final Map<String, AtomicInteger> usedIds;
```
michael-o, line 77: Why atomic? All you need is a mutuable integer, no? This component is neither singleton not in MT use.

doxia-core/src/main/java/org/apache/maven/doxia/index/IndexingSink.java
```diff
@@ -323,6 +331,26 @@ public void text( String text )
...
+    String getUniqueId( String id )
+    {
+        final String uniqueId;
+        if ( usedIds.containsKey( id ) )
+        {
+            uniqueId = id + "_" + usedIds.get( id ).incrementAndGet();
+        }
+        else
+        {
+            usedIds.put( id, new AtomicInteger() );
```
michael-o, line 348: What about `computeIfAbsent()`?

michael-o (changes requested):

doxia-core/src/main/java/org/apache/maven/doxia/index/IndexingSink.java
kwin: Which MutableInteger implementation should I use? I don't want to implement it myself and the performance drawback is IMHO neglectable here.

michael-o: Ideally from Commons Lang, but we don't have it in the class path. Since for a single class it doesn't really make sense, let's stick to it and please add a comment which depicts that we explicitly don't require the atomic semantics, but the mutuability only.

doxia-core/src/main/java/org/apache/maven/doxia/index/IndexingSink.java
michael-o: The idea is just to remove the ugly if else with is there, no then create.

michael-o: Agreed.

michael-o:
Do you actually have an example where you have found this?

doxia-core/src/main/java/org/apache/maven/doxia/index/IndexingSink.java
kwin: Done in https://github.com/apache/maven-doxia/pull/106/commits/41ad9261b547cfeda7f8cb7ac5827f18fcf0f598.

kwin:
> Do you actually have an example where you have found this?

I found it with maven-site-plugin with a markdown resource looking like this

```
<!-- MACRO{toc} -->

# Section1

## Subsection

This is subsection of section1

# Section2

## Subsection

This is subsection of section2

```

In that case the TOC is not working as the anchor towards `subsection` is not unique.

michael-o:
I see, level two heading are character-wise identical. Makes sense, then IDs are clashing.

PR #114, opened 2022-08-02
[DOXIA-668] Remove all obsolete attributes in HTML5

michael-o:
https://issues.apache.org/jira/browse/DOXIA-668

### DIFF
--- a/doxia-core/src/main/java/org/apache/maven/doxia/parser/Xhtml5BaseParser.java
+++ b/doxia-core/src/main/java/org/apache/maven/doxia/parser/Xhtml5BaseParser.java
@@ -632,7 +632,6 @@ public class Xhtml5BaseParser
         else if ( parser.getName().equals( HtmlMarkup.TABLE.toString() ) )
         {
             sink.tableRows_();
-
             sink.table_();
         }
         else if ( parser.getName().equals( HtmlMarkup.TR.toString() ) )
@@ -1044,21 +1043,11 @@ public class Xhtml5BaseParser
         }
         else
         {
-            String name = parser.getAttributeValue( null, Attribute.NAME.toString() );
-
-            if ( name != null )
+            String id = parser.getAttributeValue( null, Attribute.ID.toString() );
+            if ( id != null )
             {
-                sink.anchor( validAnchor( name ), attribs );
+                sink.anchor( validAnchor( id ), attribs );
                 isAnchor = true;
-            }
-            else
-            {
-                String id = parser.getAttributeValue( null, Attribute.ID.toString() );
-                if ( id != null )
-                {
-                    sink.anchor( validAnchor( id ), attribs );
-                    isAnchor = true;
-                }
             }
         }
     }
--- a/doxia-core/src/main/java/org/apache/maven/doxia/parser/Xhtml5BaseParser.java
+++ b/doxia-core/src/main/java/org/apache/maven/doxia/parser/Xhtml5BaseParser.java
@@ -21,6 +21,7 @@ package org.apache.maven.doxia.parser;
 
 import java.io.Reader;
 import java.util.Stack;
+import java.util.regex.Pattern;
 
 import javax.swing.text.html.HTML.Attribute;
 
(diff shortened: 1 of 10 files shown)


hboutemy:
@michael-o do we provide an example of how to use CSS instead?

and IIUC, we'll need to update our apt format reference documentation https://maven.apache.org/doxia/references/apt-format.html , for example, to show the updated behaviour (for example "+-- instead of --- draws a box around verbatim text.", or "| Centered | Left-aligned | Right-aligned  |")?

michael-o:
> @michael-o do we provide an example of how to use CSS instead?
> 
> and IIUC, we'll need to update our apt format reference documentation https://maven.apache.org/doxia/references/apt-format.html , for example, to show the updated behaviour (for example "+-- instead of --- draws a box around verbatim text.", or "| Centered | Left-aligned | Right-aligned |")?

Very good question. This I need to check. I also ran MSITE with current version of Doxia (M3) and upcoming M4 and the only difference I have seen is that some plugin output is broken due https://github.com/apache/maven-doxia/pull/114/files#diff-d01624ca9c1507eb1b1a45dbae19428e9a80912f545953340b53a0d09f015c00L114-L127

Give me some time...

michael-o:
@hboutemy Added a commit for the table grid. As far as I udnerstand `border="1"` is a combination of CSS applied to `table` and `td`/`th` so we will leave this to the skin at most. Looking into alignment...

doxia-core/src/main/java/org/apache/maven/doxia/sink/impl/Xhtml5BaseSink.java
```diff
@@ -1621,29 +1503,6 @@ private void tableCell( boolean headerRow, MutableAttributeSet attributes )
...
-                if ( attributes == null )
-                {
-                    attributes = new SinkEventAttributeSet();
-                }
-                if ( atts != null )
-                {
-                    attributes.addAttributes( atts );
-                }
-            }
-        }
```
michael-o, line 1645: Looking into this one..

doxia-core/src/main/java/org/apache/maven/doxia/parser/Xhtml5BaseParser.java
```diff
@@ -1218,26 +1212,13 @@ private void handleSectionEnd( Sink sink )
...
-        if ( "center".equals( align ) )
-        {
-            justif[0] = Sink.JUSTIFY_CENTER;
-        }
-        else if ( "right".equals( align ) )
-        {
-            justif[0] = Sink.JUSTIFY_RIGHT;
-        }
-
-        sink.tableRows( justif, grid );
```
michael-o, line 1241: This could be a fixed, pregiven CSS class name, but then it does not make sense to parse because it will be passed through.

doxia-core/src/main/java/org/apache/maven/doxia/sink/impl/Xhtml5BaseSink.java
```diff
@@ -1621,29 +1503,6 @@ private void tableCell( boolean headerRow, MutableAttributeSet attributes )
...
-            && cellJustifStack != null && !cellJustifStack.isEmpty() && getCellJustif() != null )
-        {
-            int cellCount = getCellCount();
-            if ( cellCount < getCellJustif().length )
-            {
-                Map<Integer, MutableAttributeSet> hash = new HashMap<>();
-                hash.put( Sink.JUSTIFY_CENTER, SinkEventAttributeSet.CENTER );
-                hash.put( Sink.JUSTIFY_LEFT, SinkEventAttributeSet.LEFT );
-                hash.put( Sink.JUSTIFY_RIGHT, SinkEventAttributeSet.RIGHT );
-                MutableAttributeSet atts = hash.get( getCellJustif()[cellCount] );
```
michael-o, line 1634: This can be converted to CSS classes... during parsing this is retained...

michael-o (commented):
@hboutemy left a few comments, but I don't see a perfect solution. Real 1:1 identity isn't possible.

michael-o:
@hboutemy I think this is now done and good enough. When parsing XHTML5 is is impossible to retain it on table, but on cell level `style` will be retained and the sink will print `style` when justification is supplied.

hboutemy (changes requested):
I think this change needs to be split to better understand impacts (or not) on different markup features:
- "<a name=>" replaced with "<a id=>"
- <p <tr <th align= replaced with style="text-align: "
- align on table seems the only align that is not retained at all
- table border

currently, the addition of everything is too hard to describe

michael-o:
@hboutemy Now split up into five separate commits. This should make it clear now.

michael-o:
@hboutemy Please check

hboutemy (approved):
LGTM: great work on splitting, which helps a lot understanding so many changes associated to the objective

thanks

jira-importer:
Resolve #824

PR #116, opened 2022-08-04
[DOXIA-670] Drop dead XdocSink#link(String, String) method

michael-o:
https://issues.apache.org/jira/browse/DOXIA-670

### DIFF
--- a/doxia-modules/doxia-module-xdoc/src/main/java/org/apache/maven/doxia/module/xdoc/XdocSink.java
+++ b/doxia-modules/doxia-module-xdoc/src/main/java/org/apache/maven/doxia/module/xdoc/XdocSink.java
@@ -504,31 +504,6 @@ public class XdocSink
         init();
     }
 
-    /**
-     * Adds a link with an optional target.
-     *
-     * @param name the link name.
-     * @param target the link target, may be null.
-     */
-    public void link( String name, String target )
-    {
-        if ( isHeadFlag() )
-        {
-            return;
-        }
-
-        MutableAttributeSet att = new SinkEventAttributeSet();
-
-        att.addAttribute( Attribute.HREF, HtmlTools.escapeHTML( name ) );
-
-        if ( target != null )
-        {
-            att.addAttribute( Attribute.TARGET, target );
-        }
-
-        writeStartTag( A, att );
-    }
-
     // ----------------------------------------------------------------------
     //
     // ----------------------------------------------------------------------
--- a/doxia-modules/doxia-module-xdoc/src/test/java/org/apache/maven/doxia/module/xdoc/XdocSinkTest.java
+++ b/doxia-modules/doxia-module-xdoc/src/test/java/org/apache/maven/doxia/module/xdoc/XdocSinkTest.java
@@ -20,12 +20,15 @@ package org.apache.maven.doxia.module.xdoc;
  */
 
 import org.apache.maven.doxia.sink.Sink;
+import org.apache.maven.doxia.sink.SinkEventAttributes;
 import org.apache.maven.doxia.sink.impl.AbstractSinkTest;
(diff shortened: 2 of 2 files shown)


michael-o:
Note: This change requires to drop: https://github.com/apache/maven-doxia/blob/3b1499d0572718fa4dcb7c6b775bedbe757d46c5/doxia-modules/doxia-module-xdoc/src/main/resources/xdoc-2.0.xsd#L2923-L2935 and release a new version. Alternatively, we'd translate source to `<div class="source"><pre>...`

doxia-modules/doxia-module-xdoc/src/main/java/org/apache/maven/doxia/module/xdoc/XdocSink.java
```diff
@@ -382,119 +377,6 @@ else if ( depth == SECTION_LEVEL_5 )
...
 
-    /**
-     * {@inheritDoc}
-     *
-     * @see XdocMarkup#SOURCE_TAG
-     * @see javax.swing.text.html.HTML.Tag#PRE
-     * @param attributes a {@link org.apache.maven.doxia.sink.SinkEventAttributes} object.
-     */
-    public void verbatim( SinkEventAttributes attributes )
-    {
```
michael-o, line 393: I don't have a better idea how to make it proper, but it is definitively broken with XHML5.

michael-o: I think this statement is nonsense and this method needs to be retained:
* Xdoc uses element of HTML5, but is not HTML5
* When XdocParser writes to a sink that sink will never see `<source>` anymore.

We have here a mix of custom elements and HTML5.

michael-o:
This is now a minimalist version. I will leave the table as is.

doxia-modules/doxia-module-xdoc/src/main/java/org/apache/maven/doxia/module/xdoc/XdocSink.java
```diff
@@ -505,31 +497,6 @@ public void close()
...
-        att.addAttribute( Attribute.HREF, HtmlTools.escapeHTML( name ) );
-
-        if ( target != null )
-        {
-            att.addAttribute( Attribute.TARGET, target );
-        }
-
-        writeStartTag( A, att );
-    }
-
```
hboutemy, line 532: it seems that this link(String, String) method is more than "duplicate code": it's dead code that was introduced into Doxia 1.1 and was never called outside the unit test
then I'm all in to remove this dead code and update the associated unit test, with proper Jira issue renaming

on other changes like VALIGN, that are unrelated to this dead code removal, I'd prefer have a separate commit and review to better understand

and I don't understand why there is a discussion on "source", which seems to be completely an unrelated discussion

hboutemy (changes requested):
let's rename this issue into "drop dead link(String, String) code" and have VALIGN removal in a separate issue, with discussion on source also separate

doxia-modules/doxia-module-xdoc/src/main/java/org/apache/maven/doxia/module/xdoc/XdocSink.java
michael-o: @hboutemy Done, completely reduced. Attributes are in another ticket. As for the `source`. HTML5 has now a source element which collides with Xdoc's source. That is why I decided not to fiddle with at the moment. Forget about it.

michael-o:
@hboutemy PR redone.

hboutemy (approved):
thanks for the hard work

jira-importer:
Resolve #826

PR #123, opened 2022-10-01
[DOXIA-671] Double quotes contained in markdown page are removed in H…

michael-o:
…TML output

Ideally, one would use this list in the future:
  https://html.spec.whatwg.org/multipage/named-characters.html#named-character-references


hboutemy (approved):
LGTM: it's good to have found the fix for such strange behaviour

michael-o:
The idea was actually from the Maven Xpp3 reader :-D

PR #126, opened 2022-10-31
[DOXIA-678] Add missing elements in Xhtml5BaseParser

michael-o:


michael-o:
@jdimeo have a look

jdimeo:
Looks good @michael-o ! However, how future proof is this approach? As you mentioned, there aren't a ton of people able to maintain Doxia any more and I'm a bit worried about all these specific tag lists in the code. Say 5 years from now they invent a `<vr>` tag for Virtual Reality... will that now blow up the logs with warnings if we use that in our site?

michael-o:
> Looks good @michael-o ! However, how future proof is this approach? As you mentioned, there aren't a ton of people able to maintain Doxia any more and I'm a bit worried about all these specific tag lists in the code. Say 5 years from now they invent a `<vr>` tag for Virtual Reality... will that now blow up the logs with warnings if we use that in our site?

I do understand your concerns and you are right. While I cannot look into the future, I have talked already about Doxia 3 with @hboutemy and Doxia 3 will likely complete drop the parser/sink approach it will be a thin layer about FlexMark/Asciidoctor maybe with JBake. I don't like Doxia's current complexity and it isn't as useful as we it was be 10 or 15 years ago. We don't need a Swiss Army knife yet another Pandoc anymore.

The current target of Doxia 2 is to drop luggage and remove workarounds. Also consider that I am the only one maintaining Doxia for the last couple of years...

jdimeo:
> Also consider that I am the only one maintaining Doxia for the last couple of years...
That is exactly what I was considering with my comment :-) thank you for your time and keeping it "alive"

Even in Doxia 2, why do we have to list unrecognized tags? why not just pass them through and make a log message at the `DEBUG` or `FINE` level?

michael-o:
> > Also consider that I am the only one maintaining Doxia for the last couple of years...
> 
> That is exactly what I was considering with my comment :-) thank you for your time and keeping it "alive"
> 
> Even in Doxia 2, why do we have to list unrecognized tags? why not just pass them through and make a log message at the `DEBUG` or `FINE` level?

The list is basically a hack, so to speak, for the reason provided in DOXIA-678. The Sink API does not offer all of HTML5. I can never because it is meant to be generic. The warning has been added way before me. Although I can consider putting it down to debug, it will just hide the problem under the carpet and not really solve it.

jdimeo:
I still vote for debug, because it doesn't actually cause a problem (the tags still render just fine when mixed with Markdown), but yet logging at `WARN` does cause a *huge* problem when you have over 45,000 files like I do. But this change also addresses the immediate need by adding my tags to the list... thank you!

jdimeo:
Or, if you want to keep it as a warning, maybe keep a static thread safe list of which tags have been WARNed already, and only log it once per tag. That's a lot of work but then would be really helpful at creating the list of tags that need to be addressed. Something like 

> WARN Unrecogized tag `<vr> at 8:234. Enable verbose logging to see all occurrences"

jdimeo:
... which is why I have been suggesting we just simply change the level from WARN to DEBUG :-)

michael-o:
> ... which is why I have been suggesting we just simply change the level from WARN to DEBUG :-)

Please create separate issue for this. Note that multiple parsers emit this warning.

michael-o:
@jdimeo Are we good to merge this one?

jdimeo:
Yes!

PR #133, opened 2022-12-23
[DOXIA-684] avoid flexmark-all: declare used extensions only

hboutemy:
https://issues.apache.org/jira/browse/DOXIA-684

### DIFF
--- a/doxia-modules/doxia-module-markdown/pom.xml
+++ b/doxia-modules/doxia-module-markdown/pom.xml
@@ -51,19 +51,62 @@ under the License.
     </contributor>
   </contributors>
 
+  <properties>
+    <flexmarkVersion>0.42.14</flexmarkVersion>
+  </properties>
+
   <dependencies>
     <dependency>
       <groupId>com.vladsch.flexmark</groupId>
-      <artifactId>flexmark-all</artifactId>
-      <version>0.42.14</version>
-      <exclusions>
-        <!-- exclude to PDF converter -->
-        <exclusion>
-          <groupId>com.vladsch.flexmark</groupId>
-          <artifactId>flexmark-pdf-converter</artifactId>
-        </exclusion>
-      </exclusions>
+      <artifactId>flexmark</artifactId>
+      <version>${flexmarkVersion}</version>
     </dependency>
+    <dependency>
+      <groupId>com.vladsch.flexmark</groupId>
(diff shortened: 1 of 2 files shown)


michael-o:
Nice, how did you determine that list?

hboutemy:
just used https://github.com/vsch/flexmark-java/blob/master/flexmark-all/pom.xml to know the existing artifactIds, then added the missing ones for the used classes: nothing hard, just some not funny time to find the 8 ones :)

doxia-modules/doxia-module-markdown/pom.xml
```diff
@@ -51,19 +51,57 @@ under the License.
     </contributor>
   </contributors>
 
+  <properties>
+    <flexmark.version>0.42.14</flexmark.version>
```
michael-o, line 55: I think we should use the camelCase approach for version properties everywhere now: `flexmarkVersion`

michael-o: master is at https://github.com/apache/maven-doxia/blob/master/doxia-modules/doxia-module-markdown/pom.xml#L67 `0.62.2`, 1.11.x requires a different version though

michael-o (changes requested):

michael-o:
Yet another issue:
```
[INFO] --- maven-dependency-plugin:3.1.1:analyze (default-cli) @ doxia-module-markdown ---
[WARNING] Used undeclared dependencies found:
[WARNING]    com.vladsch.flexmark:flexmark-util:jar:0.42.14:compile
```

doxia-modules/doxia-module-markdown/pom.xml
hboutemy: done the camelCase, you're right, it's what we do for parent POMs
on version, yes, the current PR is for Doxia 1.x, Doxia 2.x will get another version

michael-o: If for 1.x then then PR is wrong. IT targets master.

hboutemy:
merged in 539428bd17e649c506e4a7a549b8f8b5d1cc3c18 for Doxia 1.x and e6848f9e4b2fdeaa3378183d8b557d55fb476ad6 for Doxia 2

michael-o (commented):
LGTM now

PR #142, opened 2023-01-24
[DOXIA-691] emit multiple authors in separate tags

kwin:


doxia-modules/doxia-module-apt/src/test/java/org/apache/maven/doxia/module/apt/AptParserTest.java
```diff
@@ -486,6 +486,24 @@ public void testLiteralAnchor() throws Exception {
         assertSinkEquals(it, "link_", "sectionTitle1_", "section1_", "body_");
     }
 
+    @Test
+    public void testMultipleAuthors() throws Exception {
+        // DOXIA-691
```
michael-o, line 491: This line is redundant since we have "git blame"

doxia-modules/doxia-module-apt/src/test/resources/test/authors.apt
```diff
@@ -0,0 +1,6 @@
+ -----
+ Test DOXIA-691
```
michael-o, line 2: Sma here..

michael-o:
This needs to be updated as well: https://maven.apache.org/doxia/references/apt-format.html

michael-o (changes requested):

doxia-modules/doxia-module-apt/src/test/java/org/apache/maven/doxia/module/apt/AptParserTest.java
kwin: It is the pattern used in all test methods, so I would say conformity over redundancy

kwin:
> This needs to be updated as well: https://maven.apache.org/doxia/references/apt-format.html

What specifically? Only the way how multiple authors are emitted in XHTML/Sink methods have changed, but this has never been documented anywhere.

michael-o:
> > This needs to be updated as well: https://maven.apache.org/doxia/references/apt-format.html
> 
> What specifically? Only the way how multiple authors are emitted in XHTML/Sink methods have changed, but this has never been documented anywhere.

So the support for multiple authors in apt sources won't be documented since you have a test doc for it.

kwin:
> So the support for multiple authors in apt sources won't be documented since you have a test doc for it.

I raised https://github.com/apache/maven-doxia-site/pull/27 for the clarification.

michael-o (approved):

PR #148, opened 2023-02-19
Clarify that it's OK to use spaces for things other than indentation

elharo:
Indeed, it's necessary,

### DIFF
--- a/README.md
+++ b/README.md
@@ -56,7 +56,7 @@ There are some guidelines which will make applying PRs easier for us:
 + Make commits of logical units.
 + Respect the original code style: by using the same [codestyle][code-style],
   patches should only highlight the actual difference, not being disturbed by any formatting issues:
-  + Only use spaces for indentation.
+  + Use spaces for indentation, not tabs.
   + Create minimal diffs - disable on save actions like reformat source code or organize imports. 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -56,7 +56,7 @@ There are some guidelines which will make applying PRs easier for us:
 + Make commits of logical units.
 + Respect the original code style: by using the same [codestyle][code-style],
   patches should only highlight the actual difference, not being disturbed by any formatting issues:
-  + Only use spaces for indentation.
```
michael-o, line 59: I guess this file is in every repo...

elharo: Probably. Is there a master version somewhere from which this is derived?

michael-o: @khmarbaise Is there one?

slachiewicz: https://maven.apache.org/developers/conventions/code.html

slachiewicz: this file was added manually, then adjusted to every repo. Maybe would be good to add also recommendations/requirement to use spotless plugin (otherwise build would fail)

 IMHO this description is too long - easier maybe would be to add a link to generic instructions to out maven home page?

slachiewicz (approved):

PR #149, opened 2023-02-19
Prefer Apache commons utils to codehaus

elharo:


michael-o (changes requested):
Since there are two users now, move to parent's dep mgmt section and use then w/o version.

doxia-core/pom.xml
```diff
@@ -53,6 +53,11 @@ under the License.
       <artifactId>commons-text</artifactId>
       <version>1.10.0</version>
     </dependency>
+    <dependency>
+      <groupId>commons-io</groupId>
+      <artifactId>commons-io</artifactId>
+      <version>2.11.0</version>
+    </dependency>
```
michael-o, line 60: Wrong scope here

michael-o (changes requested):
The change should read: Commons IO, not Utils

doxia-core/pom.xml
elharo: Probably still a few to go, but with the additional files I've added  this is now needed in compile scope.

michael-o (changes requested):
The deps still must be managed.

elharo (commented):
dependencies managed

elharo (commented):
title updated

michael-o (changes requested):
Why are you adding `org.apache.commons.lang3.StringUtils` although you don't use it?

elharo:
org.apache.commons.lang3.StringUtils is used. However in this case it's a drop-in replacement for codehaus.plexus StringUtils — e.g. in line 2381 of AptParser — so only the import statement needed to change. You can see it in the whole file, just not in the diff fragment.

doxia-modules/doxia-module-xdoc/src/main/java/org/apache/maven/doxia/module/xdoc/XdocParser.java
```diff
@@ -333,7 +330,7 @@ private void handleMetaStart(XmlPullParser parser, Sink sink, SinkEventAttribute
 
     private void handleParamStart(XmlPullParser parser, Sink sink) throws MacroExecutionException {
         if (!isSecondParsing()) {
-            if (StringUtils.isNotEmpty(macroName)) {
+            if (macroName != null && !macroName.isEmpty()) {
                 String paramName = parser.getAttributeValue(null, Attribute.NAME.toString());
```
michael-o, line 334: But why remove these if StringUtils is present?

PR #152, opened 2023-03-01
[DOXIA-699] Add failing test case

kwin:


kwin:
@michael-o Can you have a look why this test fails?

michael-o:
Looking into...

michael-o (commented):
@kwin The input file is invalid because it has not been generated by Doxia 2.

kwin:
Why is it invalid? Every valid XHTML should work with the Doxia parser.

michael-o:
> Why is it invalid? Every valid XHTML should work with the Doxia parser.

Very good question, at least the diff doesn't look suspicious: https://github.com/apache/maven-doxia/commit/13138a96894b0ad341a36a069595ad5bcea0ba7c

michael-o:
Maybe this is it:
```
- <h2><a name="Feature_Summary"></a>Feature Summary</h2>
+ <h1>Feature Summary</h1>
```

Give it a shot.

kwin:
I wrote that already in the JIRA ticket: the culprit is https://github.com/apache/maven-doxia/blob/37173794270ee68894aac0e252dcb87545770952/doxia-core/src/main/java/org/apache/maven/doxia/parser/Xhtml5BaseParser.java#L631

michael-o:
But the exception is totally misleading...

michael-o:
I'd say Doxia should not try to heal the document...

kwin:
> I'd say Doxia should not try to heal the document...

Maybe that is rather up for discussion with a follow-up issue. Right now I would only want to fix the missing close element method.

michael-o:
> > I'd say Doxia should not try to heal the document...
> 
> Maybe that is rather up for discussion with a follow-up issue. Right now I would only want to fix the missing close element method.

Agree...

kwin:
The PR at #162 provides a simpler reproducer/test case of this issue.

kwin:
Superseded by #162.

jira-importer:
Resolve #836

PR #156, opened 2023-03-19
[MNG-6825] Replace Plexus IOUtil with IOUtils from commons-io

timtebeek:
For https://issues.apache.org/jira/browse/MNG-6825

This third PR targets IOUtil, which was not present in apache/maven, hence I've targeted the project that uses it the most.

Regrettably some manual fixes where needed, as the replacement commons-io IOUtils does not silently swallow IOExceptions like Plexus IOUtil did. Hence inlined the old behaviour a couple times and adopted try-with-resources a few other times.

Not entirely happy both with the process for these replacements and with the outcomes. PR opened to gather feedback on whether we would want to target the other replacements similarly, or adopt a different approach, such as more try-with-resources, or no commons-io IOUtils.

### DIFF
--- a/doxia-core/src/test/java/org/apache/maven/doxia/sink/impl/AbstractSinkTest.java
+++ b/doxia-core/src/test/java/org/apache/maven/doxia/sink/impl/AbstractSinkTest.java
@@ -22,7 +22,6 @@ import java.io.CharArrayWriter;
 import java.io.IOException;
 import java.io.Writer;
 
-import org.apache.commons.io.IOUtils;
 import org.apache.maven.doxia.AbstractModuleTest;
 import org.apache.maven.doxia.sink.Sink;
 import org.apache.maven.doxia.sink.SinkEventAttributes;
@@ -110,7 +109,6 @@ public abstract class AbstractSinkTest extends AbstractModuleTest {
             SinkTestDocument.generate(testSink);
         } finally {
             testSink.close();
-            IOUtils.close(writer);
         }
     }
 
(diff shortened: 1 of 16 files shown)


doxia-core/src/test/java/org/apache/maven/doxia/sink/impl/AbstractSinkTest.java
```diff
@@ -110,7 +110,7 @@ public final void testTestDocument() throws IOException {
             SinkTestDocument.generate(testSink);
         } finally {
             testSink.close();
-            IOUtil.close(writer);
+            IOUtils.close(writer);
```
elharo, line 113: This one might be able to use try with resources.

elharo (approved):

doxia-core/src/test/java/org/apache/maven/doxia/sink/impl/AbstractSinkTest.java
timtebeek: Looked at this briefly, but `Sink` does not implement `AutoCloseable`, which makes it slightly awkward, especially as some of the multiple `Sink#close()` implementations call out to `init()` for instance, so leaving out `testSink.close()` is also not an option. You'd then have to declare `Sink` outside the try with resources, and close that through finally, which could change the order of the `close` calls if I'm not mistaken. Any other suggestions on how to handle this here?

elharo: https://issues.apache.org/jira/browse/DOXIA-698 but until then this probably works. Does closing the sink close the Writer? If so we don't need IOUtils at all here.

The other pre-existing issue that;s weird in this test is the isXmlSink() branch on line 106. I'm not sure what that's about.

timtebeek: It's `Sink.close()` documented as that it should:

PR #159, opened 2023-03-25
TODO was done

elharo:
TODO comments are evil.


michael-o (approved):
But consider that other modules set this value explicitly. It should be ocnsistent.

PR #160, opened 2023-03-25
[DOXIA-698] Make Sink implement AutoCloseable

elharo:


michael-o (commented):
Wait a second. Who will close the writer or stream passed? Impl is required?

elharo:
That behavior doesn't change. If a Sink implementation closes the stream/writer when the sink is closed, then it will still be closed when stream/writer is autoclosed. If the Sink implementation doesn't close the  stream/writer when the sink is closed, then it won't be closed now. Nothing in this PR changes what happens when a Sink's close method is invoked. It just enables try-with-resources to be used to implicitly invoke the close method instead of explicitly calling it. This makes for slightly simpler code and makes resource leaks where the sink isn't closed at all less likely.

michael-o (approved):
Makes sense.

PR #161, opened 2023-03-30
Fix snippet reader tab and indent handling

Jotschi:
The PR addresses the following issues:

* The snippet reader did not correctly apply the indentation for snippet sources which were indented using tabs. The minIndent calculation only checked spaces.

Snippet Source:
```
[tab][tab]line
```

Old Behaviour
```
[tab][tab]line
```

New Behaviour:
```
line
```

* The snippet reader minIndent calculation was always 0 whenever the snippet contained empty newlines. Those empty lines will now correctly be handled.

Snippet Source:
```
[space][space]line1
[\n]
[space]line2
```

Old Behaviour
```
[space][space]line1
[\n]
[space]line2
```

New Behaviour:
```
line1
[\n]
[space]line2
```

### DIFF
--- a/doxia-core/src/main/java/org/apache/maven/doxia/macro/snippet/SnippetReader.java
+++ b/doxia-core/src/main/java/org/apache/maven/doxia/macro/snippet/SnippetReader.java
@@ -73,7 +73,9 @@ public class SnippetReader {
         int minIndent = minIndent(lines);
         StringBuffer result = new StringBuffer();
         for (String line : lines) {
-            result.append(line.substring(minIndent));
+            if (line.length() > minIndent) {
+                result.append(line.substring(minIndent));
+            }
             result.append(EOL);
         }
         return result;
@@ -88,6 +90,9 @@ public class SnippetReader {
     int minIndent(List<String> lines) {
         int minIndent = Integer.MAX_VALUE;
         for (String line : lines) {
+            if (StringUtils.isBlank(line)) {
+                continue;
+            }
             minIndent = Math.min(minIndent, indent(line));
         }
         return minIndent;
@@ -103,7 +108,7 @@ public class SnippetReader {
         char[] chars = line.toCharArray();
         int indent = 0;
         for (; indent < chars.length; indent++) {
-            if (chars[indent] != ' ') {
+            if (chars[indent] != ' ' && chars[indent] != '\t') {
                 break;
             }
         }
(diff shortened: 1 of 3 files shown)


michael-o:
I don't understand this change. It does not retain the original format, what does it address?

Jotschi:
It fixes the indentation behaviour when the source contains tabs. Without the fix the indentation is not correctly handled. Without the fix the source code block has too many tabs and is not rendered in a nice format.

michael-o:
> It fixes the indentation behaviour when the source contains tabs. Without the fix the indentation is not correctly handled. Without the fix the source code block has too many tabs and is not rendered in a nice format.

But looking at your examples it does remove tabs as well, no?

michael-o:
First snippet has two tabs, but expected does not have any. What do I miss?

Jotschi:
The tabs from this example 
```[tab][tab]line```
will not be removed. Thus the snippet looks ugly. The tabs should be removed.

**Comment by michael-o** [comment]
@kwin, do you understand this PR?

**Comment by Jotschi** [comment]
The snippet renderer removed unnecessary indentation. It thus moved the code block left. A code example which is highly indented will thus not include the bogus indentation.

```java
{
     {
snipped-start
             { 
                    // your code 
             }
snipped-end
      }
}
```

should result in

```java
{ 
      // your code 
}
```

and not

```java
             { 
                    // your code 
             }
```

michael-o:
But what if
```
             { 
                    // your code 
             }
```
is correct indentation?

Jotschi:
The code clearly removes this indentation if it is done by using spaces. Using tabs it is not removed and thus looks broken.
I'm okay with it if that is the desired behaviour.

michael-o:
> The code clearly removes this indentation if it is done by using spaces. Using tabs it is not removed and thus looks broken. I'm okay with it if that is the desired behaviour.

It is hard to say what is right, but I would say that a snippet should remain as-is and not modified, as ugly as it is.

Jotschi:
I think the indentation is clearly removed in order to prevent horizontal scrollbars for included snippets.

michael-o:
> I think the indentation is clearly removed in order to prevent horizontal scrollbars for included snippets.

I'd prefer to leave formatting as-is, but I am open to a snippet option which reverse-indents.

michael-o:
I'd like to perform a release this weekend. Anyone want to complete the discussion with me?

slachiewicz:
The design question this stalled on — whether snippet de-indentation should happen at all, and whether it should become an option — is now tracked separately in #1084, so it is not buried in this thread.

The defects you reported are still present on master: tab-indented snippets keep their indentation while space-indented ones do not. Leaving this open pending the direction taken in #1084.

PR #162, opened 2023-03-31
[DOXIA-699] Correctly generate sections for headings

kwin:


kwin:
As #152 was modified this simple test case shows the error in a more concise way.

michael-o:
I believe that our parsing assumptions are wrong. [This](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/Heading_Elements?retiredLocale=de) says nothing that we need a `<section>`. Looking at [this](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/section) `<h1>` isn't even wrapped in a section. I think that it is not our job to fix/reconcile bad input. Shit in, shit out. I dare to say to drop this: https://github.com/apache/maven-doxia/blob/37173794270ee68894aac0e252dcb87545770952/doxia-core/src/main/java/org/apache/maven/doxia/parser/Xhtml5BaseParser.java#L631-L636

kwin:
The problem is that the only appropriate event for `headline` in Sink API is `sectionTitle(int level, SinkEventAttributes attributes)` which explicitly states in its javadoc:

> This element is optional, but if it exists, it has to be contained, and be the first
element, within a corresponding {@link #section(int,SinkEventAttributes) section}
element of the same level..

The problem is that HTML is less strict than Sink API in this regard, therefore we have to fix HTML to generate a compliant Sink event list. Therefore we cannot get rid of the functionality to add missing sections.

michael-o:
> The problem is that the only appropriate event for `headline` in Sink API is `sectionTitle(int level, SinkEventAttributes attributes)` which explicitly states in its javadoc:
> 
> > This element is optional, but if it exists, it has to be contained, and be the first
> > element, within a corresponding {@link #section(int,SinkEventAttributes) section}
> > element of the same level..
> 
> The problem is that HTML is less strict than Sink API in this regard, therefore we have to fix HTML to generate a compliant Sink event list. Therefore we cannot get rid of the functionality to add missing sections.

We can change the API, no? That would be acceptable in a major version for me. We document the expected usage and behavior. If you don't comply: behavior will be undefined.

kwin:
> We can change the API, no?

This has quite an impact, e.g. it would no longer be possible to convert XHTML to APT (as APT has the same strict semantics). Then the logic for consolidation would need to be implemented in `AptSink` and potentially other sinks which have the same strict section semantics.
Also `TocMacro` currently only generates index entries for sections. Therefore I would say to rather not change the Sink semantics but fix the logic to add the missing sections (that won't do any harm to HTML).

michael-o:
> > We can change the API, no?
> 
> This has quite an impact, e.g. it would no longer be possible to convert XHTML to APT (as APT has the same strict section semantics). Then the logic for consolidation would need to be implemented in `AptSink` and potentially other sinks which have the same strict section semantics. Also `TocMacro` currently only generates index entries for sections. Therefore I would say to rather not change the Sink semantics but fix the logic to add the missing sections (that won't do any harm to HTML).

Accepted. Your logic is sound.

michael-o (approved):
No objections

jira-importer:
Resolve #836

PR #175, opened 2023-09-07
[DOXIA-706] Support SinkEventAttributes in Sink.text(...)

kwin:


michael-o (approved):
This makes sense to me.

PR #176, opened 2023-10-08
Add test for preserving (and not adding additional) URL encoding

kwin:
Make Sink.link(...) javadoc more HTML agnostic by just referencing the URI RFC


michael-o (approved):
Makes sense

PR #180, opened 2023-11-28
[DOXIA-722] Optionally create anchors for index entries (used in TOC  macro)

kwin:
Introduce SinkWrapper concept
Enforce anchor name uniqueness across documents

Some preliminary documentation in https://github.com/apache/maven-doxia-site/pull/30.

### DIFF
--- a/doxia-core/src/main/java/org/apache/maven/doxia/index/IndexEntry.java
+++ b/doxia-core/src/main/java/org/apache/maven/doxia/index/IndexEntry.java
@@ -19,11 +19,18 @@
 package org.apache.maven.doxia.index;
 
 import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.Collections;
 import java.util.List;
+import java.util.Objects;
+
+import org.apache.maven.doxia.markup.Markup;
+import org.apache.maven.doxia.sink.Sink;
 
 /**
- * <p>IndexEntry class.</p>
+ * Representing the index tree within a document with the most important metadata per entry.
+ * Currently this only contains entries for sections, but in the future may be extended, therefore it
+ * is recommended to use {@link #getType()} to filter out irrelevant entries.
  *
  * @author <a href="mailto:trygvis@inamo.no">Trygve Laugst&oslash;l</a>
  */
@@ -39,6 +46,11 @@ public class IndexEntry {
     private String id;
 
     /**
+     * true if there is already an anchor for this
+     */
+    private boolean hasAnchor;
+
+    /**
      * The entry title.
      */
     private String title;
@@ -48,13 +60,50 @@ public class IndexEntry {
      */
     private List<IndexEntry> childEntries = new ArrayList<>();
 
-    /**
-     * System-dependent EOL.
-     */
-    private static final String EOL = System.getProperty("line.separator");
+    public enum Type {
+        /**
+         * Used for unknown types but also for the root entry
+         */
+        UNKNOWN(),
+        SECTION_1(Sink.SECTION_LEVEL_1),
+        SECTION_2(Sink.SECTION_LEVEL_2),
+        SECTION_3(Sink.SECTION_LEVEL_3),
+        SECTION_4(Sink.SECTION_LEVEL_4),
+        SECTION_5(Sink.SECTION_LEVEL_5),
+        SECTION_6(),
+        DEFINED_TERM(),
+        FIGURE(),
+        TABLE();
+
+        private final int sectionLevel;
+
+        Type() {
+            this(-1);
+        }
+
+        Type(int sectionLevel) {
+            this.sectionLevel = sectionLevel;
+        }
+
+        static Type fromSectionLevel(int level) {
+            if (level < Sink.SECTION_LEVEL_1 || level > Sink.SECTION_LEVEL_5) {
+                throw new IllegalArgumentException("Level must be between " + Sink.SECTION_LEVEL_1 + " and "
+                        + Sink.SECTION_LEVEL_5 + " but is " + level);
+            }
+            return Arrays.stream(Type.values())
+                    .filter(t -> level == t.sectionLevel)
+                    .findAny()
+                    .orElseThrow(() -> new IllegalStateException("Could not find enum for sectionLevel " + level));
+        }
+    };
 
     /**
-     * Constructor.
+     * The type of the entry, one of the types defined by {@link IndexingSink}
+     */
+    private final Type type;
+
+    /**
+     * Constructor for root entry.
      *
      * @param newId The id. May be null.
      */
@@ -69,12 +118,24 @@ public class IndexEntry {
      * @param newId     The id. May be null.
      */
     public IndexEntry(IndexEntry newParent, String newId) {
+        this(newParent, newId, Type.UNKNOWN);
+    }
+
+    /**
+     * Constructor.
+     *
+     * @param newParent The parent. May be null.
+     * @param newId     The id. May be null.
+     * @param
+     */
+    public IndexEntry(IndexEntry newParent, String newId, Type type) {
         this.parent = newParent;
         this.id = newId;
 
         if (parent != null) {
             parent.childEntries.add(this);
         }
+        this.type = type;
     }
 
     /**
@@ -103,6 +164,34 @@ public class IndexEntry {
      */
     protected void setId(String id) {
         this.id = id;
+    }
+
+    /**
+     * Returns the type of this entry. Is one of the types defined by {@link IndexingSink}.
+     * @return the type of this entry
+     * @since 2.0.0
+     */
+    public Type getType() {
+        return type;
+    }
+
+    /** Set if the entry's id already has an anchor in the underlying document.
+     *
+     * @param hasAnchor {@true} if the id already has an anchor.
+     * @since 2.0.0
+     */
+    public void setAnchor(boolean hasAnchor) {
+        this.hasAnchor = hasAnchor;
+    }
+
+    /**
+     * Returns if the entry's id already has an anchor in the underlying document.
+     * @return {@code true} if the id already has an anchor otherwise {@code false}.
+     *
+     * @since 2.0.0
+     */
+    public boolean hasAnchor() {
+        return hasAnchor;
     }
 
     /**
@@ -266,7 +355,7 @@ public class IndexEntry {
             message.append(", title: ").append(title);
         }
 
-        message.append(EOL);
+        message.append(Markup.EOL);
 
         StringBuilder indent = new StringBuilder();
 
@@ -279,5 +368,30 @@ public class IndexEntry {
         }
 
         return message.toString();
+    }
+
+    @Override
+    public int hashCode() {
+        return Objects.hash(childEntries, hasAnchor, id, parent, title, type);
+    }
+
+    @Override
+    public boolean equals(Object obj) {
+        if (this == obj) {
+            return true;
+        }
+        if (obj == null) {
+            return false;
+        }
+        if (getClass() != obj.getClass()) {
+            return false;
+        }
+        IndexEntry other = (IndexEntry) obj;
+        return Objects.equals(childEntries, other.childEntries)
+                && hasAnchor == other.hasAnchor
+                && Objects.equals(id, other.id)
+                && Objects.equals(parent, other.parent)
+                && Objects.equals(title, other.title)
+                && type == other.type;
     }
 }
--- a/doxia-core/src/main/java/org/apache/maven/doxia/index/IndexingSink.java
+++ b/doxia-core/src/main/java/org/apache/maven/doxia/index/IndexingSink.java
@@ -23,49 +23,25 @@ import java.util.Map;
 import java.util.Stack;
 import java.util.concurrent.atomic.AtomicInteger;
 
+import org.apache.maven.doxia.index.IndexEntry.Type;
+import org.apache.maven.doxia.sink.Sink;
 import org.apache.maven.doxia.sink.SinkEventAttributes;
+import org.apache.maven.doxia.sink.impl.BufferingSinkProxyFactory;
+import org.apache.maven.doxia.sink.impl.BufferingSinkProxyFactory.BufferingSink;
 import org.apache.maven.doxia.sink.impl.SinkAdapter;
 import org.apache.maven.doxia.util.DoxiaUtils;
 
 /**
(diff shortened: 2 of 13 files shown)


doxia-core/src/main/java/org/apache/maven/doxia/sink/impl/NamesForIndexEntriesSinkWrapper.java
```diff
@@ -0,0 +1,136 @@
...
+    @Override
+    public void section5_() {
+        super.section5_();
+        onEndSection();
+    }
+
+    private void onStartSection(int level) {
+        currentLocation = currentLocation.createChildSection(level);
+        IndexEntry entry = entryByLocation.get(currentLocation);
+        if (entry != null) {
```
kwin, line 127: Probably a guard similar to https://github.com/apache/maven-doxia-sitetools/pull/58/files#diff-dabe16ceeba5dc7cb0a15e30dcb7c204a78f4763346f016ab7639ab2f59e78d4L250 needs to be introduced to at least prevent double anchors for APT format.

doxia-core/src/main/java/org/apache/maven/doxia/macro/MacroRequest.java
```diff
@@ -40,15 +40,23 @@ public class MacroRequest {
...
+     * @param currentParser a {@link org.apache.maven.doxia.parser.Parser} object (which is the parser triggering this macro).
      * @param sourceContent a {@link java.lang.String} object.
-     * @param parser a {@link org.apache.maven.doxia.parser.AbstractParser} object.
+     * @param parser a new {@link org.apache.maven.doxia.parser.AbstractParser} object acting as secondary parser.
      * @param param a {@link java.util.Map} object.
      * @param basedir a {@link java.io.File} object.
      */
-    public MacroRequest(String sourceContent, AbstractParser parser, Map<String, Object> param, File basedir) {
+    public MacroRequest(
+            Parser currentParser,
```
michael-o, line 54: This is exactly the idea I had, but rejected it because the I consider a macro something which should not push back, i.e., side effect free.

michael-o:
> > This seems like a reimplementation of RandomAccessSink: The RandomAccessSink provides the ability to create a {https://github.com/link Sink} with hooks. Why not evaluate it if it there.
> 
> I was not aware of this class, but `RandomAccessSink` is complex to grasp for me. I would rather go for reimplementing RandomAccessSink as special SinkWrapper. AFAIK RandomAccessSink does not allow to intercept sink API calls and enrich them but just does some buffering and concatenating multiple sink outputs into one effective one.

I don't see  need to modify/intercept the sink at all, a parser can do this perfectly with some buffering. `TOCMacro` does this buffering with a fresh parser as well.

> > This should not be related to the TOC macro although the TOC macro will benefit. A user can request auto anchors regardless of the TOC macro. Don't conflate them.
> 
> For me the TOC is the primary use case, but I am ok with always generating anchors and TOC macro will just use those.

Correct, TOC will use what has been generated, therefore a parser switch...

> > I think that a solution should be logically identical to org.apache.maven.doxia.parser.Parser.setEmitComments(boolean). A parser feature will intercept section title text and buffer until sectionTitle_() is hit, anchor and title (heading) are emitted.
> 
> As long as the default is "emit anchors for every section title" I am fine with adding an option to disable that. I don't think we should make that an opt-in behaviour as that would be breaking change from a user's perspective.

Well, making it default would be breaking because Apt, for example does not have implicit anchors. What I see is tht the opt-in will be enabled from the Maven Site Plugin invocation and not Doxia directly. That will cover your use case and won't break anything for others.

kwin:
> I don't see need to modify/intercept the sink at all, a parser can do this perfectly with some buffering. TOCMacro does this buffering with a fresh parser as well.

The point is that a sink wrapper can be used with all parsers, while buffering on parser level has to be implemented per parser.

michael-o:
> > I don't see need to modify/intercept the sink at all, a parser can do this perfectly with some buffering. TOCMacro does this buffering with a fresh parser as well.
> 
> The point is that a sink wrapper can be used with all parsers, while buffering on parser level has to be implemented per parser.

Are you certain about this?  Can't this be done in the `AbstractParser`?

doxia-core/src/main/java/org/apache/maven/doxia/parser/Parser.java
```diff
@@ -83,4 +85,34 @@ public interface Parser {
...
+    Collection<SinkWrapperFactory> getSinkWrapperFactories();
+
+    /**
+     * Determines whether to automatically generate anchors for each section found by {@link IndexingSink} or not.
+     * By default no anchors are generated.
+     *
+     * @param emitAnchors {@code true} to emit anchors otherwise {@code false} (the default)
+     * @since 2.0.0
+     */
+    void setEmitAnchors(boolean emitAnchors);
```
michael-o, line 110: This should be more specific: `setEmitSectionTitleAnchors`

michael-o (changes requested):

doxia-core/src/main/java/org/apache/maven/doxia/parser/Parser.java
kwin: It is potentially for all index entries (although IndexingSink currently only generates ones for sections, but there are some todos for figures, tables, ...). For sure it is not specific to to ones having a title....

kwin: Maybe `setEmitAnchorsForIndexEntries(...)`?

michael-o: You mean everything would could potentially have a number group? Then maybe `setEmitAnchorsForIndexableEntries()` and docs will carry only the information that it applies to section titles for the moment?

doxia-core/src/main/java/org/apache/maven/doxia/index/IndexEntry.java
```diff
@@ -105,6 +110,10 @@ protected void setId(String id) {
         this.id = id;
     }
 
+    public boolean hasAnchor() {
+        return hasAnchor;
+    }
+
     /**
      * Returns the title.
```
michael-o, line 118: Am I stupid, where is the setter?

doxia-core/src/main/java/org/apache/maven/doxia/index/IndexingSink.java
```diff
@@ -76,19 +80,41 @@ public class IndexingSink extends SinkAdapter {
...
     }
 
+    /**
+     * This should only be called once the sink is closed.
+     * Before that the tree might not be complete.
+     * @return the tree of entries starting from the root
+     */
+    public IndexEntry getRootEntry() {
+        return rootEntry;
+    }
```
michael-o, line 117: You see a way to throw ISE if the documented condition is not met?

doxia-core/src/main/java/org/apache/maven/doxia/parser/AbstractParser.java
```diff
@@ -189,6 +229,25 @@ protected boolean isSecondParsing() {
         return secondParsing;
     }
 
+    @Override
+    public void registerSinkWrapperFactory(SinkWrapperFactory factory) {
```
michael-o, line 233: Maybe rather `get`/`add` semantic names?

doxia-core/src/main/java/org/apache/maven/doxia/parser/SinkWrapperFactory.java
```diff
@@ -0,0 +1,43 @@
...
+     * @param sink the delegate
+     * @return a new sink wrapping the given one
+     */
+    Sink createWrapper(Sink sink);
+
+    /**
+     * Determines the order of sink wrappers. The highest ranked wrapper factory is called first.
+     * @return the rank of this factory
+     */
+    int getRank();
```
michael-o, line 42: In DI we use the term priority. Maybe that be suit here too? This making the terms consistent while `Integer.MAX_VALUE` would be the highest one?

Comment talks about order, but method name is rank...

doxia-core/src/main/java/org/apache/maven/doxia/index/IndexingSink.java
```diff
@@ -104,15 +106,21 @@ private IndexingSink(IndexEntry rootEntry, Sink delegate) {
...
     /**
      * This should only be called once the sink is closed.
      * Before that the tree might not be complete.
      * @return the tree of entries starting from the root
+     * @throws IllegalStateException in case the sink was not closed yet
      */
     public IndexEntry getRootEntry() {
+        if (!isComplete) {
+            throw new IllegalStateException(
+                    "The sink has not been closed yet, i.e. the index tree is not complete yet");
```
michael-o, line 122: This sink...

michael-o (changes requested):
One last nit

michael-o (approved):
This is fine for me now. I cannot come up with a better solution in reasonable time. Before this get merged I'd prefer that @hboutemy will also take a look as well. This is a lot of code and needs eyes. Give him some time.

doxia-core/src/main/java/org/apache/maven/doxia/index/IndexEntry.java
```diff
@@ -304,7 +310,7 @@ public String toString(int depth) {
             message.append(", title: ").append(title);
         }
 
-        message.append(EOL);
+        message.append(System.lineSeparator());
```
michael-o, line 313: Attention, this is a semantical change. All others use the idiom above and there are tests which change the separator with try/finally. This one is caches upon first call. I wouldn't change it.

michael-o (changes requested):

michael-o:
@kwin Please rename with a proper DOXIA JIRA issue. Maybe it would even make sense to split up in (a) Wrapper Factory approach and (b) Auto Indexer Feature

doxia-core/src/main/java/org/apache/maven/doxia/index/IndexEntry.java
kwin: I reverted that change now @michael-o

michael-o (changes requested):

doxia-core/src/main/java/org/apache/maven/doxia/index/IndexEntry.java
```diff
@@ -53,10 +56,41 @@ public class IndexEntry {
...
+        SECTION_3,
+        SECTION_4,
+        SECTION_5,
+        SECTION_6,
+        DEFINED_TERM,
+        FIGURE,
+        TABLE;
+
+        static Type fromSectionLevel(int level) {
+            if (level < SECTION_1.ordinal() || level > SECTION_6.ordinal()) {
```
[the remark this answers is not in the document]
kwin: Indeed, fixed in https://github.com/apache/maven-doxia/pull/180/commits/bc5b7ad641f5369ef14e4aa26529d49d01870e6e

doxia-core/src/main/java/org/apache/maven/doxia/index/IndexEntry.java
```diff
@@ -53,10 +56,41 @@ public class IndexEntry {
...
     /**
      * The type of the entry, one of the types defined by {@link IndexingSink}
      */
-    private final int type;
+    private final Type type;
+
+    /**
+     * System-dependent EOL.
+     */
+    private static final String EOL = System.getProperty("line.separator");
```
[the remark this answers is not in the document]
kwin: Because you said in https://github.com/apache/maven-doxia/pull/180#discussion_r1428911493

>  there are tests which change the separator with try/finally

Not sure which tests you referred to and how they change it. If they change via reflection we must not consolidate the constants either...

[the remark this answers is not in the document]
michael-o: See: AbstractModuleTest.

doxia-core/src/main/java/org/apache/maven/doxia/sink/impl/UniqueAnchorNamesValidatorFactory.java
```diff
@@ -0,0 +1,38 @@
...
+
+    @Override
+    public Sink createWrapper(Sink sink) {
+        return new UniqueAnchorNamesValidator(sink);
+    }
+
+    @Override
+    public int getPriority() {
+        // should come last (after potential preprocessing/modification of anchor names)
+        return -1000;
```
michael-o, line 36: Does using `Integer.MIN_VALUE` make sense?

michael-o (approved):

michael-o:
@hboutemy Are you able to review within a week or so?

slachiewicz (approved):
can we somehow check it based on our maven-site and the doxia converter?

kwin:
> can we somehow check it based on our maven-site and the doxia converter?

Doesn't affect Doxia Converter, for checking with m-site-p you need a patch for m-site-p and m-doxia-sitetools (see some WIP PRs in https://issues.apache.org/jira/browse/MSITE-1000 and https://issues.apache.org/jira/browse/DOXIASITETOOLS-322).

michael-o (changes requested):
The the actual change is perfectly fine, this needs a new JIRA issue because it does not really solve DOXIA-710 because we cannot solve the problem within the skin.
Please create a new issue and change the commit message. It can only solve external linking and interplay between TOC macro and the content itself.

kwin:
> The the actual change is perfectly fine, this needs a new JIRA issue because it does not really solve [DOXIA-710](https://issues.apache.org/jira/browse/DOXIA-710) because we cannot solve the problem within the skin. Please create a new issue and change the commit message. It can only solve external linking and interplay between TOC macro and the content itself.

Done in https://github.com/apache/maven-doxia/pull/180/commits/4518d58bd45591164b3043a46ad3db4356a48c85 associated with new JIRA issue [DOXIA-722](https://issues.apache.org/jira/browse/DOXIA-722)

michael-o:
See my {request](https://issues.apache.org/jira/browse/DOXIA-722?focusedCommentId=17803575&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-17803575) to rephrase the the issue summary/title.

kwin:
I renamed the JIRA title, the commit message should be fine already.

michael-o (approved):
Good work. Let's merge!

jira-importer:
Resolve #847

PR #182, opened 2023-12-14
[DOXIA-709] Cleanup overloaded Doxia 1.0 methods

kwin:
Centrally delegate from Doxia 1.0 to Doxia 1.1 or 2.0 equivalents in AbstractSink.
Clarify in javadoc of Sink.
Only implement Doxia 1.1/2.0 variants in Sinks.

### DIFF
--- a/doxia-core/src/main/java/org/apache/maven/doxia/sink/impl/AbstractSink.java
+++ b/doxia-core/src/main/java/org/apache/maven/doxia/sink/impl/AbstractSink.java
@@ -20,10 +20,12 @@ package org.apache.maven.doxia.sink.impl;
 
 import org.apache.maven.doxia.markup.Markup;
 import org.apache.maven.doxia.sink.Sink;
+import org.apache.maven.doxia.sink.SinkEventAttributes;
 
 /**
  * An abstract base class that defines some convenience methods for sinks.
- * Also acts as compatibility bridge for Doxia 1.0 methods which have overridden variants in Doxia > 1.0
+ * Also acts as compatibility bridge for Doxia 1.0 methods which have overloaded variants in Doxia > 1.0 (taking an additional argument {@link SinkEventAttributes}).
+ * This implementation just delegates the former to the latter with argument {@link SinkEventAttributes} being {@code null}.
  *
  * @author ltheussl
  * @author <a href="mailto:vincent.siveton@gmail.com">Vincent Siveton</a>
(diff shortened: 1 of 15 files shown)


kwin:
This allows to get rid of 1750 lines of code :-)

michael-o:
> This allows to get rid of 1750 lines of code :-)

Man, this I love!

doxia-core/src/main/java/org/apache/maven/doxia/sink/impl/AbstractSink.java
```diff
@@ -23,12 +23,337 @@
 
 /**
  * An abstract base class that defines some convenience methods for sinks.
+ * Also acts as compatibility bridge for Doxia 1.0 methods which have overridden variants in Doxia > 1.0
```
michael-o, line 26: So this is temporary?

doxia-core/src/test/java/org/apache/maven/doxia/sink/impl/SinkEventTestingSink.java
```diff
@@ -803,11 +488,6 @@ public void division(SinkEventAttributes attributes) {
...
     }
 
-    @Override
-    public void verbatim() {
-        addEvent("verbatim");
-    }
-
     @Override
     public void verbatim(SinkEventAttributes attributes) {
         addEvent("verbatim", new Object[] {attributes});
```
michael-o, line 493: These changes I do not fully understand. Now the events are gone?

michael-o (commented):
Did you run this down the chain until Maven Site Plugin? Any surprises?

doxia-core/src/main/java/org/apache/maven/doxia/sink/impl/AbstractSink.java
kwin: As long as the methods not taking SinkEventAttributes are not deprecated, this is permanent.

doxia-core/src/test/java/org/apache/maven/doxia/sink/impl/SinkEventTestingSink.java
kwin: No, the events are only added through the methods with the SinkEventAttributes arguments. The other ones just internally delegate to those (via the AbstractSink).

kwin:
> Did you run this down the chain until Maven Site Plugin? Any surprises?

No I haven't yet.

michael-o (approved):
No objections any more, I trust you here doing the right thing.

PR #184, opened 2023-12-23
[DOXIA-711] Introduce SinkWrapper and according SinkWrapperFactory

kwin:
Add a BufferingSinkProxyWrapper which buffers all calls on the underlying sink until it is flushed.


doxia-core/src/main/java/org/apache/maven/doxia/parser/SinkWrapper.java
kwin, line 1: Should we move this to another package? Maybe `o.a.m.doxia.sink.impl` is better, but this is clearly no impl details but rather class to be reused among doxia modules and potentially other consumers. Usually `impl` packages are used for packages which should not be reused outside the current JAR. So maybe `o.a.m.doxia.sink` is better, OTOH all other classes in `o.a.m.doxia.sink.impl` are probably not belonging to an impl package either, as all of them are reused. WDYT @michael-o

kwin: I moved now to the existing `o.a.m.doxia.sink.impl` although I find the `impl` package name part confusing...

doxia-core/src/main/java/org/apache/maven/doxia/sink/impl/PipelineSink.java
```diff
@@ -28,8 +28,10 @@
 
 /**
  * May be used to invoke the same method on a List of Sinks.
+ * Consider using the more flexible {@link SinkWrapper} approach which doesn't require the use of dynamic proxies.
```
kwin, line 31: Should we deprecate in favour of `SinkWrapper`?

doxia-core/src/main/java/org/apache/maven/doxia/sink/impl/RandomAccessSink.java
```diff
@@ -27,19 +27,21 @@
...
-import org.apache.maven.doxia.sink.SinkEventAttributes;
 import org.apache.maven.doxia.sink.SinkFactory;
 
 /**
  * The RandomAccessSink provides the ability to create a {@link Sink} with hooks.
  * A page can be prepared by first creating its structure and specifying the positions of these hooks.
  * After specifying the structure, the page can be filled with content from one or more models.
  * These hooks can prevent you to have to loop over the model multiple times to build the page as desired.
- *
+ * Consider using the {@link BufferingSinkProxyFactory} instead which allows to buffer on the (higher) Sink API level
```
kwin, line 37: Should we deprecate in favour of `SinkWrapper`?

doxia-core/src/main/java/org/apache/maven/doxia/sink/impl/BufferingSinkProxyFactory.java
```diff
@@ -0,0 +1,112 @@
...
+import java.lang.reflect.InvocationTargetException;
+import java.lang.reflect.Method;
+import java.lang.reflect.Proxy;
+import java.util.LinkedList;
+import java.util.Queue;
+
+import org.apache.maven.doxia.sink.Sink;
+
+/**
+ * Buffers all method calls on the wrapped Sink until the proxy sink's method {@link Sink#flush()} is called.
```
michael-o, line 31: proxied sink?

doxia-core/src/main/java/org/apache/maven/doxia/sink/impl/BufferingSinkProxyFactory.java
```diff
@@ -0,0 +1,112 @@
...
+            this.method = method;
+            this.args = args;
+        }
+
+        void invoke(Object object) {
+            try {
+                method.invoke(object, args);
+            } catch (IllegalAccessException | IllegalArgumentException | InvocationTargetException e) {
+                throw new IllegalStateException("Could not call buffered method " + method, e);
+            }
```
michael-o, line 50: What is the computational overhead by using reflection? Ignorable?

doxia-core/src/main/java/org/apache/maven/doxia/sink/impl/BufferingSinkProxyFactory.java
```diff
@@ -0,0 +1,112 @@
...
+        BufferingSinkProxy proxy = new BufferingSinkProxy(delegate);
+        return (Sink) Proxy.newProxyInstance(
+                delegate.getClass().getClassLoader(), new Class<?>[] {BufferingSink.class}, proxy);
+    }
+
+    public static BufferingSink castAsBufferingSink(Sink sink) {
+        if (sink instanceof BufferingSink) {
+            return (BufferingSink) sink;
+        } else {
+            throw new IllegalArgumentException("The given sink is no BufferingSink but a " + sink.getClass());
```
michael-o, line 104: ...is not a...

doxia-core/src/main/java/org/apache/maven/doxia/sink/impl/PipelineSink.java
michael-o: If you consider it better suited, I would trust your judgement here.

doxia-core/src/main/java/org/apache/maven/doxia/sink/impl/RandomAccessSink.java
michael-o: Same as `PipelineSink`

doxia-core/src/main/java/org/apache/maven/doxia/sink/impl/SinkWrapperFactoryComparator.java
```diff
@@ -0,0 +1,31 @@
...
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package org.apache.maven.doxia.sink.impl;
+
+import java.util.Comparator;
+
+/** Sorts the given {@link SinkWrapperFactory}s so that the one with the highest rank comes first (i.e. order by descending ranking) */
```
michael-o, line 23: Didn't we agree on the term 'priority'?

michael-o (changes requested):

doxia-core/src/main/java/org/apache/maven/doxia/parser/SinkWrapper.java
michael-o: I need to think about this...

doxia-core/src/main/java/org/apache/maven/doxia/sink/impl/BufferingSinkProxyFactory.java
kwin: According to https://stackoverflow.com/a/1856285/5155923 it should be neglectable in this case.

michael-o: Right, thanks.

doxia-core/src/main/java/org/apache/maven/doxia/sink/impl/BufferingSinkProxyFactory.java
kwin: Done

doxia-core/src/main/java/org/apache/maven/doxia/sink/impl/SinkWrapperFactoryComparator.java
kwin: Yes indeed, fixed now.

doxia-core/src/main/java/org/apache/maven/doxia/sink/impl/BufferingSinkProxyFactory.java
kwin: Rewritten in https://github.com/apache/maven-doxia/pull/184/commits/2695cb63c7b638f5a01c961f320c5a785eb192b3

doxia-core/src/main/java/org/apache/maven/doxia/sink/impl/RandomAccessSink.java
kwin: Ok, deprecated it in https://github.com/apache/maven-doxia/pull/184/commits/2695cb63c7b638f5a01c961f320c5a785eb192b3.

doxia-core/src/main/java/org/apache/maven/doxia/sink/impl/PipelineSink.java
kwin: Deprecated in https://github.com/apache/maven-doxia/pull/184/commits/2695cb63c7b638f5a01c961f320c5a785eb192b3.

doxia-core/src/main/java/org/apache/maven/doxia/parser/SinkWrapper.java
michael-o: > Should we move this to another package? Maybe `o.a.m.doxia.sink.impl` is better, but this is clearly no impl details but rather class to be reused among doxia modules and potentially other consumers. Usually `impl` packages are used for packages which should not be reused outside the current JAR. So maybe `o.a.m.doxia.sink` is better, OTOH all other classes in `o.a.m.doxia.sink.impl` are probably not belonging to an impl package either, as all of them are reused. WDYT @michael-o?

I agree with you completely. Only interfaced and package private stuff belongs here. Abstract ones and reusables don't. Though, since the rest is here and we cannot move them, I'd go with the compromise and put in into `impl` to make it consistently weird -- at least.

michael-o (approved):

PR #187, opened 2023-12-29
[DOXIA-716] Update and unify XMLReader creation and configuration

michael-o:
This closes #187

### DIFF
--- a/doxia-core/src/main/java/org/apache/maven/doxia/util/XmlValidator.java
+++ b/doxia-core/src/main/java/org/apache/maven/doxia/util/XmlValidator.java
@@ -18,24 +18,26 @@
  */
 package org.apache.maven.doxia.util;
 
-import javax.xml.XMLConstants;
+import javax.xml.parsers.ParserConfigurationException;
+import javax.xml.parsers.SAXParser;
+import javax.xml.parsers.SAXParserFactory;
 
 import java.io.IOException;
 import java.io.StringReader;
+import java.util.Locale;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
 
 import org.apache.maven.doxia.markup.XmlMarkup;
-import org.apache.maven.doxia.parser.AbstractXmlParser.CachedFileEntityResolver;
 import org.apache.maven.doxia.parser.ParseException;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
+import org.xml.sax.EntityResolver;
 import org.xml.sax.InputSource;
 import org.xml.sax.SAXException;
 import org.xml.sax.SAXParseException;
 import org.xml.sax.XMLReader;
 import org.xml.sax.helpers.DefaultHandler;
-import org.xml.sax.helpers.XMLReaderFactory;
 
 /**
  * A class to validate xml documents.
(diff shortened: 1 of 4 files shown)


doxia-core/src/test/java/org/apache/maven/doxia/xsd/AbstractXmlValidator.java
```diff
@@ -157,8 +161,17 @@ protected void setValidate(boolean validate) {
     private XMLReader getXMLReader() {
         if (xmlReader == null) {
             try {
-                xmlReader = XMLReaderFactory.createXMLReader("org.apache.xerces.parsers.SAXParser");
+                SAXParserFactory parserFactory = SAXParserFactory.newInstance();
```
kwin, line 164: Is it possible to shared the same method in AbstractXmlValidator and XmlValidator somehow? It seems the same features need to be enabled here.

kwin (approved):
LGTM

doxia-core/src/test/java/org/apache/maven/doxia/xsd/AbstractXmlValidator.java
michael-o: I bet so. Let me evaluate that later this day.

slachiewicz (approved):
looks good to me. 
While we are here maybe something for parsing and later for validation can be hardened:
https://cheatsheetseries.owasp.org/cheatsheets/XML_External_Entity_Prevention_Cheat_Sheet.html

michael-o:
> looks good to me. While we are here maybe something for parsing and later for validation can be hardened: https://cheatsheetseries.owasp.org/cheatsheets/XML_External_Entity_Prevention_Cheat_Sheet.html

This should be separate, please file a JIRA issue.

doxia-core/src/test/java/org/apache/maven/doxia/xsd/AbstractXmlValidator.java
michael-o: Done and ran through doxia-sitetools and m-site-p.

doxia-core/src/main/java/org/apache/maven/doxia/util/XmlValidator.java
```diff
@@ -65,57 +86,42 @@ public class XmlValidator {
...
 
     /**
-     * @param hasDtdAndXsd to flag the <code>ErrorHandler</code>.
      * @return an xmlReader instance.
      * @throws SAXException if any
+     * @throws ParserConfigurationException
      */
-    private XMLReader getXmlReader(boolean hasDtdAndXsd) throws SAXException {
+    public XMLReader getXmlReader() throws SAXException, ParserConfigurationException {
         if (xmlReader == null) {
```
kwin, line 101: I am not sure if caching is wise here, as SAXParser/SAXParserFactory is not not thread-safe AFAIK (https://docs.oracle.com/cd/E17802_01/webservices/webservices/docs/1.5/api/javax/xml/parsers/SAXParserFactory.html). At least this should be documented in the javadoc that the returned XMLReader needs to be synchronized somehow to be used among multiple threads.

michael-o: No objections, but that would be a behavioral change and should be a separate issue.

PR #190, opened 2023-12-30
[DOXIA-717] Update org.codehaus.plexus:plexus-testing to 1.3.0

slachiewicz:


pom.xml
```diff
@@ -188,8 +188,8 @@ under the License.
       <dependency>
         <groupId>com.google.inject</groupId>
         <artifactId>guice</artifactId>
-        <version>4.2.3</version>
-        <classifier>no_aop</classifier>
+        <version>6.0.0</version>
+        <scope>test</scope>
```
michael-o, line 192: How does this relate to Plexus Testing?

slachiewicz: Plexus testing uses it. With the old version -  tests will fail

michael-o: Alright, I see. Please leave a comment in the commit message so that this information does not get lost.

michael-o (approved):

jira-importer:
Resolve #968

PR #194, opened 2024-01-12
[DOXIA-723] Expose document location in Sink

kwin:


doxia-core/src/main/java/org/apache/maven/doxia/parser/AbstractXmlParser.java
```diff
@@ -194,6 +215,7 @@ protected SinkEventAttributeSet getAttributesFromParser(XmlPullParser parser) {
      * @throws org.apache.maven.doxia.macro.MacroExecutionException if there's a problem executing a macro
      */
     private void parseXml(XmlPullParser parser, Sink sink) throws XmlPullParserException, MacroExecutionException {
+        sink.setDocumentLocator(new XmlPullParserLocator(parser, null));
```
kwin, line 218: @michael-o Is the 3rd argument of https://maven.apache.org/doxia/doxia/doxia-core/apidocs/org/apache/maven/doxia/parser/Parser.html#parse-java.io.Reader-org.apache.maven.doxia.sink.Sink-java.lang.String- supposed to contain the file path? The javadoc is a bit fuzzy here....

kwin: https://issues.apache.org/jira/browse/DOXIA-224

michael-o: See: https://issues.apache.org/jira/browse/DOXIA-224 and https://github.com/apache/maven-doxia/commit/70642f0fd1dbc7d0fb7780d347b87e1fd9b2d1b0. It seems to be a logical reference omitting the format. I think it is similar to `generator` and `document`: https://github.com/apache/maven-doxia-sitetools/blob/5fe4a4c5359e6a23b78f385e15f77767cadaee99/doxia-site-renderer/src/main/java/org/apache/maven/doxia/siterenderer/DocumentRenderingContext.java#L66

See how I have been using it throughout.

michael-o: You cannot and must not assume that a parse consumes from a file, it can be any source satisfying the `Reader` interface.

michael-o:
I will check this on the weekend...

doxia-core/src/main/java/org/apache/maven/doxia/sink/impl/AbstractLocator.java
```diff
@@ -0,0 +1,69 @@
...
+    }
+
+    /**
+     * Creates a string with line/column information. Inspired by
+     * {@code o.a.m.model.building.ModelProblemUtils.formatLocation(...)}.
+     *
+     * @param locator The locator must not be {@code null}.
+     * @return The formatted location or an empty string if unknown, never {@code null}.
+     */
+    public static String formatLocation(Locator locator) {
```
michael-o, line 52: I believe that both `formatLocation` and `getLogPrefix` are misplaced here. This class is expected to contain on the four datapoints, formatting should be done else where, by the client cod, e.g. `AbstractParser` or so.

doxia-core/src/main/java/org/apache/maven/doxia/sink/impl/AbstractSink.java
```diff
@@ -420,4 +424,17 @@ protected static String unifyEOLs(String text) {
...
         // nop
     }
+
+    @Override
+    public void setDocumentLocator(Locator locator) {
+        this.locator = locator;
+    }
+
+    @Override
+    public Locator getDocumentLocator() {
```
michael-o, line 434: It is weird that one needs a locator for a sink...

michael-o (changes requested):

doxia-core/src/main/java/org/apache/maven/doxia/sink/impl/AbstractLocator.java
kwin: This is used in `Sink` (not in `Parser`), therefore the only option would be to move to `AbstractSink`.

kwin: Moved in https://github.com/apache/maven-doxia/pull/194/commits/ed298b000c4f95578755e8233cd0a4fb59dc2e8c.

doxia-core/src/main/java/org/apache/maven/doxia/sink/impl/AbstractSink.java
michael-o: I understand, I have no better idea to solve this...

doxia-sink-api/src/main/java/org/apache/maven/doxia/sink/Locator.java
```diff
@@ -0,0 +1,48 @@
...
+     *
+     * @return the column number for the sink event
+     */
+    int getColumnNumber();
+
+    /**
+     * Returns the underlying source reference (for file based documents a relative file path, otherwise an arbitrary string or {@code null}).
+     * @return the source for the sink event (may be {@code null})
+     */
+    String getReference();
```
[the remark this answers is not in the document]
kwin: Fixed in https://github.com/apache/maven-doxia/pull/194/commits/5aca45f0c859e16d5907f2690410b2aec002c4ac.

michael-o (approved):
LGTM now, nice work.

jira-importer:
Resolve #972

PR #195, opened 2024-01-13
[DOXIA-724] Fix order of sink wrappers in pipeline

kwin:


doxia-core/src/main/java/org/apache/maven/doxia/parser/Parser.java
```diff
@@ -95,11 +95,12 @@ public interface Parser {
...
     /**
-     *
-     * @return all sink wrapper factories in the correct order (both registered automatically and manually)
+     * Returns all sink wrapper factories (both registered automatically and manually). The collection is ordered in a way that
+     * the factories having the lowest priority come first (i.e. in reverse order).
+     * @return all sink wrapper factories in the reverse order
      * @since 2.0.0
      */
-    Collection<SinkWrapperFactory> getSinkWrapperFactories();
+    List<SinkWrapperFactory> getSinkWrapperFactories();
```
michael-o, line 103: Is is necessary to change the signature here?

kwin: I think it was a mistake exposing that via `Parser`. This is clearly SPI only and therefore only belongs to `AbstractParser`. There it should be a protected method. `List` as return value makes sense for usages which try to retrieve the last element only.

michael-o: The drop it from `Parser` altogether, no?

michael-o (approved):

PR #197, opened 2024-01-18
[DOXIA-727] Fix escaping inside code spans

kwin:
Code cleanup to keep track of surrounding element contexts Remove redundant flags
Get rid of deprecated classes

### DIFF
--- a/doxia-core/src/test/java/org/apache/maven/doxia/sink/impl/AbstractSinkTest.java
+++ b/doxia-core/src/test/java/org/apache/maven/doxia/sink/impl/AbstractSinkTest.java
@@ -613,7 +613,7 @@ public abstract class AbstractSinkTest extends AbstractModuleTest {
         if (isXmlSink()) {
             assertThat(wrapXml(actual), isIdenticalTo(wrapXml(expected)));
         } else {
-            assertEquals(actual, expected);
+            assertEquals(expected, actual);
         }
     }
 
@@ -632,7 +632,7 @@ public abstract class AbstractSinkTest extends AbstractModuleTest {
         if (isXmlSink()) {
             assertThat(wrapXml(actual), isIdenticalTo(wrapXml(expected)));
         } else {
-            assertEquals(actual, expected);
+            assertEquals(expected, actual);
         }
     }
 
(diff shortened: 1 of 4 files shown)


doxia-core/src/test/java/org/apache/maven/doxia/sink/impl/AbstractSinkTest.java
```diff
@@ -632,7 +632,7 @@ public void testFigureFromUrl() {
         if (isXmlSink()) {
             assertThat(wrapXml(actual), isIdenticalTo(wrapXml(expected)));
         } else {
-            assertEquals(actual, expected);
+            assertEquals(expected, actual);
```
michael-o, line 635: These are quite embrassing...

doxia-modules/doxia-module-markdown/src/main/java/org/apache/maven/doxia/module/markdown/MarkdownSink.java
```diff
@@ -64,56 +66,81 @@ public class MarkdownSink extends AbstractTextSink implements MarkdownMarkup {
...
 
-    /**  listStyles. */
-    private final Stack<String> listStyles;
+    /** Keep track of end markup for inline events. */
+    protected Queue<Queue<String>> inlineStack = Collections.asLifoQueue(new LinkedList<>());
 
-    /** Keep track of the closing tags for inline events. */
-    protected Stack<List<String>> inlineStack = new Stack<>();
+    /** The context of the surrounding elements as stack (LIFO) */
+    protected Queue<ElementContext> elementContextStack = Collections.asLifoQueue(new LinkedList<>());
```
michael-o, line 91: Stupid question: Why not use a stack directly?

michael-o (approved):
Nice amount of work, no objections here.

slachiewicz (approved):

doxia-modules/doxia-module-markdown/src/main/java/org/apache/maven/doxia/module/markdown/MarkdownSink.java
kwin: It says in its javadoc
> A more complete and consistent set of LIFO stack operations is provided by the [Deque](https://docs.oracle.com/javase/8/docs/api/java/util/Deque.html) interface and its implementations, which should be used in preference to this class.

michael-o: I see, thanks!

jira-importer:
Resolve #855

PR #198, opened 2024-01-30
[DOXIA-728] Support blockquotes in MarkdownSink

kwin:
Streamline nested block handling
Always collapse subsequent newlines and blank lines

### DIFF
--- a/doxia-modules/doxia-module-markdown/src/main/java/org/apache/maven/doxia/module/markdown/MarkdownMarkup.java
+++ b/doxia-modules/doxia-module-markdown/src/main/java/org/apache/maven/doxia/module/markdown/MarkdownMarkup.java
@@ -77,11 +77,11 @@ public interface MarkdownMarkup extends TextMarkup {
     /** Syntax for the link start: "](" */
     String LINK_START_2_MARKUP = "](";
 
-    /** Syntax for the ordered list item: '1.' */
-    String LIST_ORDERED_ITEM_START_MARKUP = "1.";
+    /** Syntax for the ordered list item: '1. ' */
+    String LIST_ORDERED_ITEM_START_MARKUP = "1. ";
 
-    /** Syntax for the unordered list item: "-" */
-    String LIST_UNORDERED_ITEM_START_MARKUP = "-";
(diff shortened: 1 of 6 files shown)


doxia-modules/doxia-module-markdown/src/main/java/org/apache/maven/doxia/module/markdown/MarkdownMarkup.java
```diff
@@ -77,11 +77,11 @@ public interface MarkdownMarkup extends TextMarkup {
...
 
-    /** Syntax for the ordered list item: '1.' */
-    String LIST_ORDERED_ITEM_START_MARKUP = "1.";
+    /** Syntax for the ordered list item: '1. ' */
+    String LIST_ORDERED_ITEM_START_MARKUP = "1. ";
 
-    /** Syntax for the unordered list item: "-" */
-    String LIST_UNORDERED_ITEM_START_MARKUP = "-";
+    /** Syntax for the unordered list item: "- " */
+    String LIST_UNORDERED_ITEM_START_MARKUP = "- ";
```
michael-o, line 84: This is rather a fix, right?

michael-o (approved):
I don't have any objections, I trust your work here.

doxia-modules/doxia-module-markdown/src/main/java/org/apache/maven/doxia/module/markdown/MarkdownMarkup.java
kwin: Previously this was combined each time with a space, but it never occurs without it, so it makes sense to include the space in the constant.

kwin:
There is currently an issue on Windows I still need to investigate.

doxia-modules/doxia-module-markdown/src/main/java/org/apache/maven/doxia/module/markdown/MarkdownMarkup.java
michael-o: Makes sense.

jira-importer:
Resolve #854

PR #202, opened 2024-03-12
[DOXIA-731] Simplify HTML markup emitted from Sink.verbatim

kwin:
Use either `<pre>` or `<pre><code>`.
Add parser tests for verbatim content (for all parsers)

### DIFF
--- a/doxia-core/src/test/java/org/apache/maven/doxia/parser/AbstractParserTest.java
+++ b/doxia-core/src/test/java/org/apache/maven/doxia/parser/AbstractParserTest.java
@@ -23,12 +23,14 @@ import java.io.Reader;
 import java.io.StringWriter;
 import java.io.Writer;
 import java.util.Iterator;
+import java.util.ListIterator;
 
 import org.apache.maven.doxia.AbstractModuleTest;
 import org.apache.maven.doxia.sink.Sink;
 import org.apache.maven.doxia.sink.SinkEventAttributes;
 import org.apache.maven.doxia.sink.impl.SinkEventAttributeSet;
 import org.apache.maven.doxia.sink.impl.SinkEventElement;
+import org.apache.maven.doxia.sink.impl.SinkEventTestingSink;
 import org.apache.maven.doxia.sink.impl.SinkWrapper;
 import org.apache.maven.doxia.sink.impl.SinkWrapperFactory;
 import org.apache.maven.doxia.sink.impl.TextSink;
@@ -36,7 +38,10 @@ import org.apache.maven.doxia.sink.impl.WellformednessCheckingSink;
 import org.junit.jupiter.api.Assertions;
 import org.junit.jupiter.api.Test;
 
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertFalse;
 import static org.junit.jupiter.api.Assertions.assertTrue;
+import static org.junit.jupiter.api.Assumptions.assumeTrue;
 
 /**
  * Test the parsing of sample input files.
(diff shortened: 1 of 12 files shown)


doxia-core/src/test/java/org/apache/maven/doxia/parser/AbstractParserTest.java
```diff
@@ -146,6 +156,91 @@ public final void testSinkWrapper() throws ParseException, IOException {
...
+    protected abstract String getVerbatimSource();
+
+    /**
+     * Test a verbatim block (no code) given through {@link #getVerbatimSource()}
+     * @throws ParseException
+     */
+    @Test
+    public void testVerbatim() throws ParseException {
+        String source = getVerbatimSource();
+        assumeTrue(source != null, "parser does not support simple verbatim text");
```
michael-o, line 190: `assertNotNull()` or similiar?

doxia-modules/doxia-module-markdown/src/test/java/org/apache/maven/doxia/module/markdown/MarkdownParserTest.java
```diff
@@ -798,4 +800,24 @@ public void testQuoteVsApostrophe() throws Exception {
...
+    }
+
+    @Override
+    protected void assertEventSuffix(Iterator<SinkEventElement> eventIterator) {
+        assertSinkEquals(eventIterator, "body_");
+    }
+
+    @Override
+    protected String getVerbatimSource() {
+        return null; // not supported in MD
```
michael-o, line 816: Indeed, CommonMark does not support this. Eventually a link to https://spec.commonmark.org/0.31.2/#fenced-code-blocks might make sense?

doxia-core/src/test/java/org/apache/maven/doxia/parser/AbstractParserTest.java
kwin: I like `assume` because then it is visually highlighted in the test result that the test was skipped (instead falsely saying it passed).

michael-o: ... and https://junit.org/junit4/javadoc/4.12/org/junit/Assume.html#assumeNotNull(java.lang.Object...)?

kwin: No longer part of JUnit5: https://junit.org/junit5/docs/5.10.2/api/org.junit.jupiter.api/org/junit/jupiter/api/Assumptions.html

michael-o: Interesting...

kwin:
> Can you test it down the line to Sitetools and Maven Site Plugin?

It works fine now together with https://github.com/apache/maven-fluido-skin/pull/58.

doxia-modules/doxia-module-markdown/src/test/java/org/apache/maven/doxia/module/markdown/MarkdownParserTest.java
kwin: Not sure fenced code blocks are related. There is also https://spec.commonmark.org/0.31.2/#indented-code-block and both really refer to code only as there is no verbatim without code semantics.

michael-o: This isn't what a meant. I meant just add the link I have provided and explain why we return null since this is not a missing/broken test.

michael-o (approved):
Except for the comment I am happy/fine with this change.

PR #204, opened 2024-03-14
[DOXIA-734] Restore handling of XHTML 1.0 obsolete attributes

kwin:
@michael-o This is what I have in mind, at least for the most important obsolete attributes

### DIFF
--- a/doxia-core/src/main/java/org/apache/maven/doxia/parser/Xhtml1BaseParser.java
+++ b/doxia-core/src/main/java/org/apache/maven/doxia/parser/Xhtml1BaseParser.java
@@ -1,0 +1,180 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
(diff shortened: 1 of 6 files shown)


doxia-core/src/main/java/org/apache/maven/doxia/parser/Xhtml1BaseParser.java
```diff
@@ -0,0 +1,159 @@
...
+    private static final class AttributeMapping {
+        private final String sourceName;
+        private final String targetName;
+        private final UnaryOperator<String> valueMapper;
+        private final MergeSemantics mergeSemantics;
+
+        enum MergeSemantics {
+            OVERWRITE,
+            IGNORE,
+            PREPEND
```
michael-o, line 49: Do we need append as well?

kwin: Let's add once we have a use case

michael-o (commented):
Very nice

kwin:
> Does it make sense to log a warning to tell the user that the attribute should be migrated?

I don't think, because the user is complying with the spec while using XHTML5 only attributes/elements (while supported) is not specified for XDoc 2.0. We should rather deprecate XDoc usage alltogether (but this is separate from this issue).

PR #214, opened 2024-05-16
[DOXIA-737] [REGRESSION] Making EmitAnchorsForIndexableEntries by def…

michael-o:
…ault does not extract document titles anymore

This closes #214

### DIFF
--- a/doxia-core/src/main/java/org/apache/maven/doxia/index/IndexingSink.java
+++ b/doxia-core/src/main/java/org/apache/maven/doxia/index/IndexingSink.java
@@ -143,30 +143,30 @@ public class IndexingSink extends org.apache.maven.doxia.sink.impl.SinkWrapper {
     public void text(String text, SinkEventAttributes attributes) {
         if (isTitle) {
             rootEntry.setTitle(text);
-            return;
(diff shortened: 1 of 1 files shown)


kwin (approved):
Thanks for fixing this. Wondering though why there is no IT/UT having caught that. Probably time to add one…

kwin:
BTW: There is no need for git commit messages which close the PR.. This is only useful when referencing a GitHub issue.

michael-o:
> BTW: There is no need for git commit messages which close the PR.. This is only useful when referencing a GitHub issue.

I know, I do it on purpose to establish the connection between the commit and the PR when browsing `git log` from the terminal. Metadata.

michael-o:
> Thanks for fixing this. Wondering though why there is no IT/UT having caught that. Probably time to add one…

Do you have an idea from the top of your head? Took me some time to tackle this. Don't whether it should be in Doxia or somewhere later this test.

kwin:
The test should be in Doxia. But I won’t have time to look into this the next two weeks. Should not block merging though

jira-importer:
Resolve #864

PR #223, opened 2024-08-16
[DOXIA-739] Disable source code highlighters for verbatim blocks for …

michael-o:
…Markdown

This closes #223

### DIFF
--- a/doxia-modules/doxia-module-markdown/src/main/java/org/apache/maven/doxia/module/markdown/MarkdownParser.java
+++ b/doxia-modules/doxia-module-markdown/src/main/java/org/apache/maven/doxia/module/markdown/MarkdownParser.java
@@ -149,6 +149,7 @@ public class MarkdownParser extends AbstractTextParser implements TextMarkup {
         flexmarkOptions.set(HtmlRenderer.HTML_BLOCK_OPEN_TAG_EOL, false);
         flexmarkOptions.set(HtmlRenderer.HTML_BLOCK_CLOSE_TAG_EOL, false);
         flexmarkOptions.set(HtmlRenderer.MAX_TRAILING_BLANK_LINES, -1);
+        flexmarkOptions.set(HtmlRenderer.FENCED_CODE_NO_LANGUAGE_CLASS, "nohighlight nocode");
 
         // Build the Markdown parser
         FLEXMARK_PARSER =
--- a/doxia-core/src/test/java/org/apache/maven/doxia/parser/AbstractParserTest.java
+++ b/doxia-core/src/test/java/org/apache/maven/doxia/parser/AbstractParserTest.java
@@ -228,7 +228,12 @@ public abstract class AbstractParserTest extends AbstractModuleTest {
         boolean isInlineCode;
         if (atts.isEmpty()) {
             isInlineCode = true;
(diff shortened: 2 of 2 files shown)


doxia-modules/doxia-module-markdown/src/main/java/org/apache/maven/doxia/module/markdown/MarkdownParser.java
```diff
@@ -149,6 +149,7 @@ public class MarkdownParser extends AbstractTextParser implements TextMarkup {
         flexmarkOptions.set(HtmlRenderer.HTML_BLOCK_OPEN_TAG_EOL, false);
         flexmarkOptions.set(HtmlRenderer.HTML_BLOCK_CLOSE_TAG_EOL, false);
         flexmarkOptions.set(HtmlRenderer.MAX_TRAILING_BLANK_LINES, -1);
+        flexmarkOptions.set(HtmlRenderer.FENCED_CODE_NO_LANGUAGE_CLASS, "nohighlight nocode");
```
kwin, line 152: Is this in line with how other parsers behave?

kwin: I don't see this considered in Fluido: https://github.com/search?q=repo%3Aapache%2Fmaven-fluido-skin+nohighlight&type=code or https://github.com/search?q=repo%3Aapache%2Fmaven-fluido-skin+nocode&type=code either...

michael-o: Yes and no. Markdown produces always `<pre><code>` while others can separate. In the Markdown case a highlighter will also highlight content which shall not be highlighted, but only be prefomatted. Both prettify and highlight.js will ignore content having these classes.

michael-o: > I don't see this considered in Fluido: https://github.com/search?q=repo%3Aapache%2Fmaven-fluido-skin+nohighlight&type=code or https://github.com/search?q=repo%3Aapache%2Fmaven-fluido-skin+nocode&type=code either...

I don't that this is necessary, it does not have to know about the semantics of the language beneath. The Markdown parser will tell whether the fenced block is plaintext or a specific language.

kwin: Well it is, as you pointed out, but only implicitly via `prettify`.

kwin (approved):

doxia-modules/doxia-module-markdown/src/main/java/org/apache/maven/doxia/module/markdown/MarkdownParser.java
michael-o: > Well it is, as you pointed out, but only implicitly via `prettify`.

Correct. `nocode` for prettify and `nohighlight` for highlight.js. I don't see a portable way to solve this with `<code>` and with Markdown there is not otherway unless we hook into the generator of Flexmark. I don't know whether that would be wise.

jira-importer:
Resolve #865

PR #226, opened 2024-09-02
Upgrade Plexus Utils and XML

michael-o:


michael-o:
@basilcrow I think you are after this, no?

michael-o:
@basil

basil:
Yes, this sounds like the cause of the Enforcer warnings with a Maven 3 baseline. Based on https://codehaus-plexus.github.io/plexus-utils/

> keep `plexus-xml` to 3 if you want Maven 3 compatibility, as `plexus-xml` 4 works only in Maven 4

cstamas:
I tend to keep all Maven-3 "thing" deps at p-u 3... and Maven-4 "things" at p-u 4 + p-x

@gnodet

michael-o:
> I tend to keep all Maven-3 "thing" deps at p-u 3... and Maven-4 "things" at p-u 4 + p-x
> 
> @gnodet

Note that we have now done the opposite everywhere for Maven 3.  Means, two deps.

gnodet:
> > I tend to keep all Maven-3 "thing" deps at p-u 3... and Maven-4 "things" at p-u 4 + p-x
> 
> > 
> 
> > @gnodet
> 
> 
> 
> Note that we have now done the opposite everywhere for Maven 3.  Means, two deps.

Maven 3: p-u 4 and p-x 3
Maven 4: p-u 4 and p-x 4

michael-o:
Any objections here still?

PR #227, opened 2024-09-07
Expose no-op sink

pzygielo:


pzygielo:
I would use it for this: https://issues.apache.org/jira/browse/MPMD-403?focusedCommentId=17880071&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-17880071

cstamas (approved):
makes sense

michael-o:
Before I consent this I will take a look at the issue with PMPD. This should not have happened. In standalone mode a preset skin should be used.

michael-o:
@pzygielo Are we good to close this one?

PR #228, opened 2024-09-17
[DOXIA-740] Add test for non-consecutive headings leading to incomplete

kwin:
output

### DIFF
--- a/doxia-core/src/main/java/org/apache/maven/doxia/index/IndexingSink.java
+++ b/doxia-core/src/main/java/org/apache/maven/doxia/index/IndexingSink.java
@@ -166,9 +166,9 @@ public class IndexingSink extends org.apache.maven.doxia.sink.impl.SinkWrapper {
                     // append text to current entry
                     IndexEntry entry = stack.lastElement();
 
-                    String title;
-                    if (entry.getTitle() != null) {
-                        title = entry.getTitle() + text;
+                    String title = entry.getTitle();
+                    if (title != null) {
+                        title += text;
                     } else {
                         title = text;
                     }
--- a/doxia-core/src/main/java/org/apache/maven/doxia/index/IndexingSink.java
+++ b/doxia-core/src/main/java/org/apache/maven/doxia/index/IndexingSink.java
@@ -54,8 +54,14 @@ public class IndexingSink extends org.apache.maven.doxia.sink.impl.SinkWrapper {
 
     private final IndexEntry rootEntry;
 
+    /** Is {@code true} once the sink has been closed. */
     private boolean isComplete;
+
     private boolean isTitle;
+
+    /** Is {@code true} if the sink is currently populating entry data (i.e. metadata about the current entry is not completely captured yet) */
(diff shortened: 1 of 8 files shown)


doxia-core/src/main/java/org/apache/maven/doxia/macro/toc/TocMacro.java
```diff
@@ -112,9 +113,13 @@ public void execute(Sink sink, MacroRequest request) throws MacroExecutionExcept
             tocSink.close();
         }
 
-        IndexEntry index = tocSink.getRootEntry();
+        writeTocForIndexEntry(sink, getAttributesFromMap(request.getParameters()), tocSink.getRootEntry());
+    }
+
+    void writeTocForIndexEntry(Sink sink, SinkEventAttributes listAttributes, IndexEntry rootEntry) {
```
kwin, line 119: This still needs some work to
a) only consider entries for type section
b) skip empty entries (not having an id)

kwin:
@michael-o Feel free to already have a look, although this need to consider some more edge cases (see comments inline).

doxia-core/src/main/java/org/apache/maven/doxia/macro/toc/TocMacro.java
kwin: Done now.

doxia-core/src/main/java/org/apache/maven/doxia/index/IndexingSink.java
```diff
@@ -150,14 +158,20 @@ public void text(String text, SinkEventAttributes attributes) {
...
+                case SECTION_6:
                     // -----------------------------------------------------------------------
                     // Sanitize the id. The most important step is to remove any blanks
                     // -----------------------------------------------------------------------
 
                     // append text to current entry
                     IndexEntry entry = stack.lastElement();
 
-                    String title = entry.getTitle() + text;
+                    String title;
```
michael-o, line 169: Why not `title = entry.getTitle()` and then `title += text`/`title = text`?

michael-o (changes requested):

doxia-core/src/main/java/org/apache/maven/doxia/index/IndexingSink.java
kwin: Because entry.getTitle() might return null

michael-o: I mean for the if else clause:
```
String title = entry.getTitle();
                    if (title != null) {
                        title += text;
                    } else {
                        title = text;
                    }
```

kwin: Done in https://github.com/apache/maven-doxia/pull/228/commits/bd1e61ea7705af4a6a6f19ad28355a5c6807c343.

michael-o (approved):

jira-importer:
Resolve #866

PR #236, opened 2024-10-12
[DOXIA-746] Sink API: add method for block comment

kwin:
Add tests for consecutive comments


doxia-sink-api/src/main/java/org/apache/maven/doxia/sink/Sink.java
```diff
@@ -1724,6 +1724,17 @@ public interface Sink extends AutoCloseable {
...
+    /**
+     * Add a comment. The default implementation will just call {@link #comment(String)}.
+     *
+     * @param comment The comment to write.
+     * @param isBlockComment If <code>true</code> this is a block comment, i.e. nothing should follow on the same line
+     * @since 2.1
+     */
+    default void comment(String comment, boolean isBlockComment) {
+        comment(comment);
+    }
```
michael-o, line 1736: I would rather name it the way it is: `void blockComment`

michael-o (changes requested):

doxia-sink-api/src/main/java/org/apache/maven/doxia/sink/Sink.java
kwin: I am still thinking about a more suitable name, because this only affects the suffix, but not the prefix of the comment (i.e. in case of `true` the comment is not necessarily the only thing in the line). Any good idea here?

kwin: Probably `endsWithLineBreak` is more accurate here.

doxia-core/src/main/java/org/apache/maven/doxia/sink/impl/Xhtml5BaseSink.java
```diff
@@ -1444,34 +1444,44 @@ public void rawText(String text) {
...
+        final StringBuilder buffer = new StringBuilder(comment.length() + 7);
 
-            write(buffer.toString());
+        buffer.append(LESS_THAN).append(BANG).append(MINUS).append(MINUS);
+        buffer.append(comment);
+        buffer.append(MINUS).append(MINUS).append(GREATER_THAN);
+        if (endsWithLineBreak) {
+            buffer.append(EOL);
         }
+        return buffer.toString();
```
michael-o, line 1484: Do you want to normalize line endings in the actual comments as well?

kwin: This happens during `Xhtml5BaseSink.write(....)`, but not in the MarkdownSink equivalent `writeUnescaped(....)`. Let's leave this for later, only very remotely related to this issue.

jira-importer:
Resolve #742

PR #238, opened 2024-10-15
[DOXIA-749] Correctly indent and separate blocks inside list items

kwin:
Block elements should be surrounded by blank lines inside list items to be compliant with more MD parsers

### DIFF
--- a/doxia-modules/doxia-module-markdown/src/main/java/org/apache/maven/doxia/module/markdown/MarkdownSink.java
+++ b/doxia-modules/doxia-module-markdown/src/main/java/org/apache/maven/doxia/module/markdown/MarkdownSink.java
@@ -873,9 +873,6 @@ public class MarkdownSink extends AbstractTextSink implements MarkdownMarkup {
 
     @Override
     public void text(String text, SinkEventAttributes attributes) {
-        if (text.equals("\n")) {
-            return;
-        }
         if (attributes != null) {
             inline(attributes);
         }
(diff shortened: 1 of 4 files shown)


doxia-modules/doxia-module-markdown/src/main/java/org/apache/maven/doxia/module/markdown/MarkdownSink.java
```diff
@@ -859,6 +873,9 @@ public void nonBreakingSpace() {
 
     @Override
     public void text(String text, SinkEventAttributes attributes) {
+        if (text.equals("\n")) {
```
michael-o, line 876: Is this portable POSIX/Windows?

kwin: Good catch, this is a leftover from some debugging, is not supposed to be in here.

kwin: It turned out that the HTML parser often emits `text("\n")`without any semantical meaning. It becomes semantical if prefixing that with spaces, though. Therefore added a workaround to not prefix empty line breaks.

doxia-modules/doxia-module-markdown/src/test/java/org/apache/maven/doxia/module/markdown/MarkdownSinkTest.java
```diff
@@ -513,4 +510,24 @@ public void testHeadingAfterInlineElement() {
...
     }
+
+    @Test
+    public void testMultilineVerbatimSourceAfterListItem() {
+        try (final Sink sink = getSink()) {
+            sink.list();
+            sink.listItem();
+            sink.text("Before");
+            sink.verbatim(SinkEventAttributeSet.SOURCE);
+            sink.text("codeline1\ncodeline2");
```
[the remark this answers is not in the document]
kwin: This is supposed to be a multiline comment, `\n` by EOL now.

PR #239, opened 2024-10-15
[DOXIA-750] Properly apply inlines inside HTML blocks

kwin:


michael-o (approved):
No objection. Maybe @hboutemy

PR #240, opened 2024-10-19
[DOXIA-751] Linked inline code must be emitted in right order

kwin:
Introduce buffer stack to be able to buffer for each context separately. Refactoring of buffer handling


jira-importer:
Resolve #870

PR #241, opened 2024-10-22
[DOXIA-753] Do not end lists with a blank line

kwin:
Otherwise every nested list is implicitly "loose" instead of "tight"

### DIFF
--- a/doxia-modules/doxia-module-markdown/src/main/java/org/apache/maven/doxia/module/markdown/MarkdownSink.java
+++ b/doxia-modules/doxia-module-markdown/src/main/java/org/apache/maven/doxia/module/markdown/MarkdownSink.java
@@ -448,7 +448,7 @@ public class MarkdownSink extends AbstractTextSink implements MarkdownMarkup {
 
     @Override
     public void list_() {
-        ensureBlankLine();
+        ensureBeginningOfLine();
     }
 
     @Override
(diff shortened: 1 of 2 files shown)


michael-o:
What about non-list blocks/elements after a list? Is there a new line then?

kwin:
> What about non-list blocks/elements after a list? Is there a new line then?

Good point, will check what CommonsMark says about it. Most probably the safest option is to always add a blank line after the outermost `list_()`

kwin:
According to https://spec.commonmark.org/0.31.2/#blank-lines blank lines never have any semantical meaning except within lists. However certain parsers think differently (especially inside block elements in lists, which require a blank line in flexmark with the options we use in Doxia). Compare with https://spec.commonmark.org/0.31.2/#why-is-a-spec-needed-:

> Is a blank line needed before a block quote or heading? Most implementations do not require the blank line. However, this can lead to unexpected results in hard-wrapped text, and also to ambiguities in parsing (note that some implementations put the heading inside the blockquote, while others do not). (John Gruber has also spoken [in favor of requiring the blank lines](https://web.archive.org/web/20170611172104/http://article.gmane.org/gmane.text.markdown.general/2146).)

> Is a blank line needed before an indented code block? (Markdown.pl requires it, but this is not mentioned in the documentation, and some implementations do not require it.)

michael-o (approved):

PR #245, opened 2024-11-04
[DOXIA-757] Don't strip leading "#" from link names

kwin:
Markdown is fundamentally different with anchor links from APT. Never try to escape link names as according to Sink API those always must be valid URIs.


michael-o (approved):

jira-importer:
Resolve #973

PR #248, opened 2024-11-13
[DOXIA-760] Clarify table justification semantics and introduce new

kwin:
"JUSTIFY_DEFAULT" alignment


michael-o (approved):

doxia-core/src/main/java/org/apache/maven/doxia/sink/impl/Xhtml5BaseSink.java
```diff
@@ -1112,6 +1109,24 @@ && getCellJustif() != null) {
         }
     }
 
+    private static String getStyleForTableJustification(int justification) {
```
michael-o, line 1112: The actual code isn't just necessarify for tables only, it could apply everywhere logically, no?

kwin: The constant defining the justification are only referring to tables according to their comments. Also nowhere else there is a Sink API taking into for justification…

michael-o: Right

michael-o (approved):

PR #251, opened 2024-12-12
[DOXIA-764] Trim left the text inside <source>

kwin:
It is block element in XDoc but potentially an inline element in the Sink (XHTML emits `<pre><code>`)

### DIFF
--- a/doxia-modules/doxia-module-xdoc/src/main/java/org/apache/maven/doxia/module/xdoc/XdocParser.java
+++ b/doxia-modules/doxia-module-xdoc/src/main/java/org/apache/maven/doxia/module/xdoc/XdocParser.java
@@ -31,6 +31,7 @@ import java.util.LinkedHashMap;
 import java.util.Map;
 
 import org.apache.commons.io.IOUtils;
+import org.apache.commons.lang3.StringUtils;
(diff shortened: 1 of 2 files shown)


doxia-modules/doxia-module-xdoc/src/main/java/org/apache/maven/doxia/module/xdoc/XdocParser.java
```diff
@@ -241,6 +246,23 @@ protected void handleEndTag(XmlPullParser parser, Sink sink)
         isEmptyElement = false;
     }
 
+    /** {@inheritDoc} */
+    protected void handleText(XmlPullParser parser, Sink sink) throws XmlPullParserException {
```
kwin, line 250: after having a 2nd thought, this rather seems to be an issue with the XhtmlSink which does not properly trim text following verbatim (with source decoration), as https://maven.apache.org/doxia/doxia/doxia-sink-api/apidocs/org/apache/maven/doxia/sink/Sink.html#verbatim-- is defined as block element (so I think assuming that its content is trimmed seems to be valid). WDYT @michael-o ?

kwin:
Superseded by #251

PR #255, opened 2025-01-02
[DOXIA-770] Only escape when necessary

kwin:
There are still cases where escaping wouldn't be necessary but checking if those require escaping is too much effort.

### DIFF
--- a/doxia-modules/doxia-module-markdown/src/main/java/org/apache/maven/doxia/module/markdown/LastTwoLinesBufferingWriter.java
+++ b/doxia-modules/doxia-module-markdown/src/main/java/org/apache/maven/doxia/module/markdown/LastTwoLinesBufferingWriter.java
@@ -91,4 +91,8 @@ public class LastTwoLinesBufferingWriter extends Writer {
     public void close() throws IOException {
         out.close();
     }
+
+    public boolean isAfterDigit() {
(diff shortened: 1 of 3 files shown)


doxia-modules/doxia-module-markdown/src/main/java/org/apache/maven/doxia/module/markdown/LastTwoLinesBufferingWriter.java
```diff
@@ -91,4 +91,8 @@ public void flush() throws IOException {
     public void close() throws IOException {
         out.close();
     }
+
+    public boolean isAfterDigit() {
+        return currentLine.length() > 1 && Character.isDigit(currentLine.charAt(currentLine.length() - 1));
+    }
```
michael-o, line 97: This also covers non-European digits. Is that expected?

kwin: The markdown spec is not very clear in this regard and also different markdown parses behave differently, therefore it is safer to also consider those as digits (i.e. escape rather too often than miss one time).

michael-o (approved):

jira-importer:
Resolve #884

PR #267, opened 2025-05-27
Enable Github Issues

Bukama:
* Release was there
* Milestone created


hboutemy:
@Bukama please apply the usual PR template approach: it has been applied on 80% of our repositories, too late to change now
we'll have that discussion in a separate time, let's not conflate topics

hboutemy:
FYI https://github.com/apache/maven-gh-actions-shared/issues/165 created to track the template simplification feature request

PR #270, opened 2025-06-19
Upgrade parent pom to 45

slachiewicz:


slachiewicz:
@slawekjaranowski site is failing. Any hint ?

slawekjaranowski:
@slachiewicz as we have a dependency to `test-jar` it is not resolver from current session during site build ... especially in forked reports ....

We move snapshot repositories to profile in parent, so we can use it ...

PR #1087, opened 2026-08-24
Close self-closing non-void HTML tags in Markdown parsing

### LINKED ISSUE #1086: Markdown parser leaves self-closing non-void HTML tags unclosed, swallowing following content
A self-closing HTML tag for a non-void element (e.g. `<object .../>`, hand-embedded in a Markdown source or produced by tooling) is rendered by `doxia-module-markdown` as an unclosed tag, and everything that follows in the same block gets nested inside it instead of being a sibling.

This is what broke https://maven.apache.org/scm.html: apache/maven-site#1645 (fixed there by hboutemy's apache/maven-site#1652, closing the `<object>` explicitly by hand).

### Reproduction

Markdown input:

```markdown
Heading
=======

<p><object type="image/svg+xml" data="x.svg"/></p>

Text after the object.
```

Converting markdown → xhtml with `doxia-converter` (2.1.0):

```
$ java -jar doxia-converter-shaded.jar -in fixture.md -from markdown -out fixture.xhtml -to xhtml
```

Output:

```html
<body><section>
<h1>Heading</h1>

<p><object type="image/svg+xml" data="x.svg">
<p></p>

<p>Text after the object.</p></object></section></body></html>
```

The `<object>` is never closed. "Text after the object." ends up nested inside it, along with the empty `<p></p>` from the source paragraph break; the real `</object>` close only appears right before `</section>`. In a browser, an `<object>` element's content is only shown as fallback when the object itself fails to load, so everything after the tag silently disappears whenever the referenced resource loads successfully — exactly the symptom on scm.html.

### Root cause

`MarkdownParser.toXhtml()` re-serializes flexmark's raw-HTML passthrough by parsing it with Jsoup and calling `outputSettings().syntax(Document.OutputSettings.Syntax.xml)`. Jsoup follows the HTML5 spec correctly here: a trailing `/` on a non-void element's start tag is a no-op, not a self-close. `<br/>`, ``, `<hr/>` etc. are fine because those *are* void elements, but `<object/>`, `<div/>`, `<span/>`, `<a/>`, ... are not, so the parser is left with a genuinely unclosed tag and nests all following sibling content as its children until the end of the containing block, where Jsoup then emits the missing close tag.

Verified with jsoup's own `Tag.valueOf(name).isSelfClosing()`: true for `img`/`br`/`hr`/`input`/`param`, false for `object`/`div`/`p`/`span`/`a`/`script`/`svg`.

Note: this is distinct from `doxia-converter`'s xdoc→markdown conversion, which doesn't reproduce this particular symptom — it drops a raw `<object>` from xdoc entirely (`Unknown Sink event 'object', ignoring!`), the same way it already drops `<map>`/`<area>` image maps. The unclosed-tag defect is purely in how `doxia-module-markdown` parses Markdown that already contains such a self-closing tag, however it got there.

*This issue was created with AI assistance.*

slachiewicz:
Closes self-closing non-void HTML tags (e.g. `<object .../>`) before Jsoup re-serializes flexmark's raw-HTML passthrough, so they no longer swallow the content that follows as unclosed-tag fallback children.

Fixes #1086

Verified: `mvn test` on `doxia-module-markdown` → all green (new test fails without the fix, passes with it).

*This change was created with AI assistance.*


michael-o:
Does this cover the case if the tag spans multiple lines? Which can happen...

slachiewicz:
Yes. Verified by running the pattern rather than reading it:

| input | result |
| --- | --- |
| `<object type="image/svg+xml" data="x.svg"/>` | closed |
| `<object` + newline + `type="…"` + newline + `data="x.svg"/>` | closed |
| `<object data="a` + newline + `b"/>` (newline inside a quoted value) | closed |
| `<br/>`, and `` | left alone, correctly |
| `<object data="a"/><object data="b"/>` | both closed |

Multi-line works because the pattern contains no `.`, so DOTALL never comes into it. The
attribute group is built from `"[^"]*"`, `'[^']*'` and `[^"'/>]`, and in Java all three match a
newline, as does the `\s*` before `/>`.

Code blocks are safe for a separate reason worth recording: flexmark escapes their content, so
`&lt;object/&gt;` never matches and a literal example in a fenced block is not rewritten.

Testing it did turn up one case it misses, though — an unquoted attribute value containing a
slash:

    <object data=x/y.svg/>   ->   unchanged

`/` is excluded from every branch of the attribute group, so the match can never reach the
trailing `/>`. HTML permits unquoted attribute values and flexmark passes raw HTML through
verbatim, so that input still hits the original bug. Narrower than the multi-line case you asked
about, but it is the real hole. Happy to either handle it or record it as a known limitation with
a test — your call which.

PR #1088, opened 2026-08-24
Render unknown Sink events as raw HTML in MarkdownSink

slachiewicz:
`MarkdownSink.unknown()` logged a warning and silently dropped any Sink event it had no dedicated handling for, taking the content it carried with it (e.g. `<object>` embeds, image maps). It now delegates tag reconstruction to `Xhtml5BaseSink.unknown()` and emits the raw HTML tag instead, restoring that content in the converted output. `<thead>`/`<tbody>`/`<tfoot>` stay excluded, since they carry no content of their own and Doxia's table sink API already captures the structure they wrap.

Aligned with #1087 (`xdoc-selfclosing-raw-html`): a self-closing, non-void tag (e.g. `<object/>`) is never rendered self-closed, since HTML5 ignores the trailing slash on non-void elements and would otherwise swallow all following content as fallback children. It is written as an explicitly closed pair instead.

Root-cause fix for apache/maven-doxia-converter#159.

Verified: end-to-end conversion of the issue's xdoc fixture (`<object type="image/svg+xml" data="x.svg"/>`) through a locally built `doxia-converter` now yields `<object type="image/svg+xml" data="x.svg"></object>` in the Markdown output instead of silently vanishing.

*This change was created with AI assistance.*

### DIFF
--- a/doxia-modules/doxia-module-markdown/src/main/java/org/apache/maven/doxia/module/markdown/MarkdownSink.java
+++ b/doxia-modules/doxia-module-markdown/src/main/java/org/apache/maven/doxia/module/markdown/MarkdownSink.java
@@ -38,6 +38,7 @@ import org.apache.maven.doxia.sink.impl.Xhtml5BaseSink;
 import org.apache.maven.doxia.util.DoxiaStringUtils;
 import org.apache.maven.doxia.util.DoxiaUtils;
 import org.apache.maven.doxia.util.HtmlTools;
+import org.jsoup.parser.Tag;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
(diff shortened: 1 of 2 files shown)


michael-o:
Is the behavior of `#unknown()` defined in the API or left to the implementor, read undefined/impl specific?

slachiewicz:
Left to the implementor, explicitly. `Sink#unknown` says:

> Adds an unknown event. This may be used by parsers to notify a general Sink about an event
> that doesn't fit into any event defined by the Sink API. Depending on the parameters, a Sink
> may decide whether or not to process the event, emit it as raw text, as a comment, log it, etc.

"emit it as raw text, as a comment, log it" is an explicit grant of latitude, and the existing
implementations already use all of it: `SinkAdapter` no-ops, `AptSink` logs and ignores,
`Xhtml5BaseSink` reconstructs the tag. So moving `MarkdownSink` from log-and-drop to emitting the
tag is a choice the contract allows rather than one it constrains.

`MarkdownSink extends Xhtml5BaseSink`, so this PR delegates to that implementation rather than
inventing its own interpretation, and only diverges where HTML5 forces it to (a self-closed
non-void tag).

PR #1091, opened 2026-09-01
Make SinkEventAttributeSet.hashCode ignore resolveParent (#1075)

### LINKED ISSUE #1075: SinkEventAttributeSet.equals and hashCode disagree about resolveParent
### Describe the bug

`SinkEventAttributeSet.hashCode()` folds in the resolve parent:

```java
public int hashCode() {
    final int parentHash = (resolveParent == null ? 0 : resolveParent.hashCode());
    return attribs.hashCode() + parentHash;
}
```

but `equals(Object)` delegates to `isEqual(AttributeSet)`, which compares only the attribute count and containment and never looks at the resolve parent. Two sets with the same own attributes but different resolve parents are therefore equal with different hash codes, which breaks the `equals`/`hashCode` contract and makes the class unsound as a key in any hash-based collection once `setResolveParent` has been used.

Raised by a review bot on #1073 and left alone there on purpose, since that PR only moves the class and changing equality semantics inside a move would have hidden a behaviour change in a diff that otherwise has none.

### Impact

Currently theoretical as far as I can see: `setResolveParent` is not called anywhere in Doxia itself, nor in the report plugins that use this class — only in the class's own unit test. So this is a latent trap rather than a live bug, which is also why it is worth fixing deliberately rather than in passing.

### Possible directions

Whichever way it goes, the two methods need to agree:

- make `hashCode()` ignore the resolve parent, matching the existing `equals`. Least disruptive, and arguably right given `AttributeSet.isEqual` is defined in terms of attributes; or
- make `equals` consider the resolve parent, matching the existing `hashCode`. Stricter, but changes when two sets compare equal, so it is the riskier of the two.

Note `equals` and `isEqual` are not required to mean the same thing — `isEqual` comes from the Swing `AttributeSet` contract — so redefining `equals` without touching `isEqual` is available if that is preferred.

### Affected version

2.1.0, and every earlier 1.x/2.x release; the code is unchanged since the class was introduced.

tanvir-ux:
`SinkEventAttributeSet.equals()` already delegates to `isEqual()`, which compares only own attributes. `hashCode()` was adding the resolve parent, so two equal sets could have different hash codes.

I dropped the parent from `hashCode()` so it matches existing equality. `getAttribute()` still walks the parent. This is the least disruptive option from #1075. I did not change equals semantics.

`checkHashCode` now expects `setResolveParent` not to change the hash. I also added a contract test for two sets with the same own attributes and different parents.

This is independent of #1073 (the class move). Happy to rebase if that lands first.

Fixes #1075

Following this checklist to help us incorporate your
contribution quickly and easily:

- [x] Your pull request should address just one issue, without pulling in other changes.
- [x] Write a pull request description that is detailed enough to understand what the pull request does, how, and why.
- [x] Each commit in the pull request should have a meaningful subject line and body.
  Note that commits might be squashed by a maintainer on merge.
- [x] Write unit tests that match behavioral changes, where the tests fail if the changes to the runtime are not applied.
  This may not always be possible but is a best-practice.
- [x] Run `mvn verify` to make sure basic checks pass.
  A more thorough check will be performed on your pull request automatically.
- [ ] You have run the integration tests successfully (`mvn -Prun-its verify`).

I ran `mvn -pl doxia-core -am test -Dtest=SinkEventAttributeSetTest -Dsurefire.failIfNoSpecifiedTests=false` locally (18 tests, green, including checkstyle/spotless/rat on those modules). I did not run `-Prun-its`.

- [x] I hereby declare this contribution to be licenced under the [Apache License Version 2.0, January 2004](http://www.apache.org/licenses/LICENSE-2.0)
- [ ] In any other case, please file an [Apache Individual Contributor License Agreement](https://www.apache.org/licenses/icla.pdf).

### DIFF
--- a/doxia-sink-api/src/main/java/org/apache/maven/doxia/sink/SinkEventAttributeSet.java
+++ b/doxia-sink-api/src/main/java/org/apache/maven/doxia/sink/SinkEventAttributeSet.java
@@ -356,9 +356,7 @@ public class SinkEventAttributeSet implements SinkEventAttributes, Cloneable {
 
     @Override
     public int hashCode() {
-        final int parentHash = (resolveParent == null ? 0 : resolveParent.hashCode());
-
-        return attribs.hashCode() + parentHash;
+        return attribs.hashCode();
     }
 
     @Override
--- a/doxia-sink-api/src/test/java/org/apache/maven/doxia/sink/SinkEventAttributeSetTest.java
+++ b/doxia-sink-api/src/test/java/org/apache/maven/doxia/sink/SinkEventAttributeSetTest.java
@@ -259,7 +259,20 @@ class SinkEventAttributeSetTest {
         oldValue = newValue;
         sinkEventAttributeSet.setResolveParent(SinkEventAttributeSet.CENTER);
         newValue = sinkEventAttributeSet.hashCode();
-        assertNotEquals(oldValue, newValue);
+        assertEquals(oldValue, newValue);
+    }
+
+    /**
+     * Own attributes decide equality; resolveParent must not break the hashCode contract.
+     */
+    @Test
+    void hashCodeAgreesWithEqualsWhenResolveParentDiffers() {
+        SinkEventAttributeSet a = new SinkEventAttributeSet(SinkEventAttributeSet.BOLD);
+        SinkEventAttributeSet b = new SinkEventAttributeSet(SinkEventAttributeSet.BOLD);
+        a.setResolveParent(SinkEventAttributeSet.CENTER);
+        b.setResolveParent(SinkEventAttributeSet.LEFT);
+        assertEquals(a, b);
+        assertEquals(a.hashCode(), b.hashCode());
     }
 
     /**


slachiewicz:
#1073 landed as e83c5a2b and moved both files this PR touches, so it now conflicts.

The code is unchanged, only relocated:

- `doxia-sink-api/src/main/java/org/apache/maven/doxia/sink/SinkEventAttributeSet.java` —
  `hashCode()` is at line 358, still adding `parentHash`
- `doxia-sink-api/src/test/java/org/apache/maven/doxia/sink/SinkEventAttributeSetTest.java`

`doxia-core` keeps a deprecated subclass that declares neither `equals` nor `hashCode`, so
the fix still applies once, in the new location. `SinkEventAttributeSetCompatibilityTest`
in `doxia-core` pins unrelated behaviour and needs no change.

The move was in flight when you opened this — apologies for the churn.

tanvir-ux:
Rebased onto master after #1073. Same hashCode change, now in `doxia-sink-api`.

slachiewicz (approved):

